PR #5, opened 2022-08-30
Add some explanations and policy about patched versions

michaelpj:
I included the "extra version components" version of the policy, let's not merge this until we find out if that actually works!


andreabedini:
Ops, well. Still in time to revert.

PR #22, opened 2022-09-15
Make sure we have gnutar in scope

### LINKED ISSUE #17: scripts/add-from-github.sh -- requires gnutar
The add-from-github script depends on some gnutar specific wildcard behaviour and fails on run on macOS.

andreabedini:
The add-from-github script depends on some gnutar specific wildcard behaviour and otherwise fails on run on macOS.

Closes #17.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,13 +11,14 @@
       (system:
         let pkgs = nixpkgs.legacyPackages.${system}; in
         {
-          packages.default = pkgs.buildEnv {
+          packages.default = with pkgs; buildEnv {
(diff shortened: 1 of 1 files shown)


angerman:
LGTM :D Though it's really more an issue if someone tries to run the command just by invoking it from the scripts folder on macOS, without having nix on hand.

andreabedini:
> LGTM :D Though it's really more an issue if someone tries to run the command just by invoking it from the scripts folder on macOS, without having nix on hand.

¯\_(ツ)_/¯ :smile:

PR #23, opened 2022-09-15
Do not rely on the existence of repo branch

andreabedini:


angerman:
Don't you want to switch to an orphan branch if there is no pre-existing repo branch? Something like
```
git switch --orphan repo
```

andreabedini:
`git commit-tree` is very low level and doesn't need a working directory. I am aware this git-fu is very complicated but I just haven't found a better way. Maybe once can use `git subtree` or `git worktree`.

PR #26, opened 2022-09-15
add-from-github script can force versions

### LINKED ISSUE #18: scripts/add-from-github.sh -- add revision field
It would be nice if one could do something like:

```
./scripts/add-from-github.sh https://github.com/input-output-hk/nothunks 45d066c4f47400e645bda7dc0ffe5070bd2baf42 -rev 3
```
and have this produce `$version.0.0.0.3`.

andreabedini:
Closes #18

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -22,11 +22,9 @@ while getopts ${optstring} arg; do
       usage
       ;;
     r)
-      echo "Revision ${OPTARG}"
       REVISION="${OPTARG}"
       ;;
     v)
-      echo "Version ${OPTARG}"
       VERSION="${OPTARG}"
       ;;
     :)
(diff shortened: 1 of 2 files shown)


scripts/add-from-github.sh
```diff
@@ -3,26 +3,78 @@
 set -o errexit
 set -o pipefail
 
+function usage {
+  echo "Usage $(basename "$0") [-r REVISION] [-v VERSION] REPO_URL REV [SUBDIRS...]"
+  echo
+  echo "        -r REVISION     adds .0.0.0.0.REVISION to the package version"
```
angerman, line 9: 0.0.0.0.$REVISION? I thought it was 0.0.0.$REVISION? is it 3 or 4 zeros?

andreabedini: The README says four zeros.
> The scheme that we typically use is to take the existing version number, add four zero components and then a patch version, e.g. 1.2.3.4.0.0.0.0.1.

angerman:
❤️

PR #35, opened 2022-09-29
Update metadata to use foliage github backend

andreabedini:
This updates everything (metadata and script) to use foliage github backend. This is so that foliage knows the details about the github repo and can use it for other tihngs (e.g. webpages or adding metadata in the cabal files).

I checked this does not cause any extra changes. On my computer only the signatures change (obviously has I don't have the same keys). Most importantly the hashes of the sdists do not change. Please test again yourself :-)

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -61,12 +61,15 @@ fi
 
 render_meta() {
   local TIMESTAMP=$1
-  local URL=$2
-  local SUBDIR=$3
-  local FORCE_VERSION=$4
+  local REPO_URL=$2
+  local REPO_REV=$3
+  local SUBDIR=$4
+  local FORCE_VERSION=$5
+
+  local REPO=${REPO_URL#https://github.com/}
 
   echo "timestamp = $TIMESTAMP"
-  echo "url = '$URL'"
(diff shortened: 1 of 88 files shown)


scripts/add-from-github.sh
```diff
@@ -61,12 +61,15 @@ fi
...
   local TIMESTAMP=$1
-  local URL=$2
-  local SUBDIR=$3
-  local FORCE_VERSION=$4
+  local REPO_URL=$2
+  local REPO_REV=$3
+  local SUBDIR=$4
+  local FORCE_VERSION=$5
+
+  local REPO=${REPO_URL#https://github.com/}
```
michaelpj, line 69: what happens if this doesn't match?

andreabedini:
Note that the check passes because I am not changing any timestamp. Nevertheless I could have replaced the packages by changing the url. This is something we should improve upon. Perhaps by adding integrity hashes to the urls.

michaelpj (approved):
Nice!

scripts/add-from-github.sh
andreabedini: I added a check.

PR #37, opened 2022-10-10
Don't put a slash at the end of the url

michaelpj:
Not needed, and it makes it easier for people to get the wrong URL when constructing the `inputMap` which makes things not work.


andreabedini (approved):
Oh man, I was sure I remembered cabal somehow needed it. But I checked again with cabal 3.6 and cabal 3.8 and it does not. :+1:

PR #41, opened 2022-10-11
Add some policy on not requiring extra build configuration

michaelpj:
WDYT @andreabedini ? It would be cool to check this, but for now I think we just have to ask people not to do it.


andreabedini (approved):
Looks good to me. I am definitely in favour.

PR #45, opened 2022-10-11
Fix the flake to actually include foliage

michaelpj:


michaelpj:
Oh huh, I actually didn't realise that you could do `nix shell` like that, it does work. Nonetheless, I think this way is a bit more idiomatic: people expect `nix develop` to get you the shell when working with flakes.

PR #59, opened 2022-10-17
Node 1.35.3

disassembler:
This breaks history, but is needed because the current listed revisions aren't from the latest node release.


andreabedini:
I am not sure we need to break the index history for this. Note that you can release older versions of a package (e.g. 0.0.9.0). Of all these packages we need to figure out if they are newer or older that the existing versions and release them as 0.0.9.0 or 0.2.0.0 (I’m on mobile I might be getting the version numbers wrong but I hope you get the point).

PR #60, opened 2022-10-19
Add a note about active-repositories

michaelpj:
WDYT @andreabedini ? Maybe we should even suggest that people _do_ add this? I realised that a difference it makes is that including it explicitly lets us say that the project should only use Hackage and CHaP, whereas the implicit version will also use any other repositories the user has configured globally... but that's quite an edge case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,20 @@ The `index-state` for the package repository can also be pinned in as usual:
 index-state: cardano-haskell-packages 2022-08-25T00:00:00Z
 ```
 
+You _can_ use the `active-repositories` stanza if you like.
+This explicitly tells cabal which repositories to use and with which precedence.
+You don't need to set this, because cabal implicitly does the right thing (CHaP will take 
+precedence, since it is defined in the project file).
+
+```
+-- note: later = higher precedence
+active-repositories: hackage.haskell.org, cardano-haskell-packages
+```
+
+IMPORTANT: do _not_ use the `:override` modifier on CHaP. 
+This is in principle a good idea, but since we add patched Hackage packages to CHaP, it 
+would prevent cabal from taking later versions of those packages from Hackage, which is not what we want.
+
 ### ... with haskell.nix
 
 To use CHaP with `haskell.nix`, do the following:


andreabedini:
It is good to mention this but I don't think we should really suggest it. What would be the problem of having other repositories available globally?

README.md
```diff
@@ -61,6 +61,20 @@ The `index-state` for the package repository can also be pinned in as usual:
 index-state: cardano-haskell-packages 2022-08-25T00:00:00Z
 ```
 
+You _can_ use the `active-repositories` stanza if you like.
```
andreabedini, line 64:
```suggestion
If you want to state explicitly the repositories that should be considered for your project,  you can use the `active-repositories` stanza.
```

README.md
```diff
@@ -61,6 +61,20 @@ The `index-state` for the package repository can also be pinned in as usual:
...
+You don't need to set this, because cabal implicitly does the right thing (CHaP will take 
+precedence, since it is defined in the project file).
+
+```
+-- note: later = higher precedence
+active-repositories: hackage.haskell.org, cardano-haskell-packages
+```
+
+IMPORTANT: do _not_ use the `:override` modifier on CHaP. 
+This is in principle a good idea, but since we add patched Hackage packages to CHaP, it
```
andreabedini, line 75:
```suggestion
The `:override` modifier tells cabal that CHaP is the sole source for any package it has available; while the default behaviour would be to consider, for each package, all the versions across all active repositories.
While this might sounds like a good idea, this
```

README.md
```diff
@@ -61,6 +61,20 @@ The `index-state` for the package repository can also be pinned in as usual:
...
+precedence, since it is defined in the project file).
+
+```
+-- note: later = higher precedence
+active-repositories: hackage.haskell.org, cardano-haskell-packages
+```
+
+IMPORTANT: do _not_ use the `:override` modifier on CHaP. 
+This is in principle a good idea, but since we add patched Hackage packages to CHaP, it 
+would prevent cabal from taking later versions of those packages from Hackage, which is not what we want.
```
andreabedini, line 76:
```suggestion
would prevent cabal from taking later versions of those packages we patched from Hackage.
```

michaelpj:
I guess I wanted to say something in case people find out about `active-repositories` and `:override`, but maybe that's unnecessarily pre-emptive.

> What would be the problem of having other repositories available globally?

If you _did_ you could get a different build plan, which could mess things up.

andreabedini:
> If you did you could get a different build plan, which could mess things up.

I am not sure. Once we have bounds in place, the build plan is up to the users and this should be safe. Are you looking at this form a security prospective or from a devx one?

PR #65, opened 2022-10-21
Add hydra-cardano-api 0.7.0

ch1bo:


michaelpj:
Based on your comments on slack, this won't actually build with packages from CHaP, since it relies on additional patches to `cardano-api`, right?

ch1bo:
> Based on your comments on slack, this won't actually build with packages from CHaP, since it relies on additional patches to `cardano-api`, right?

Ah yes, you're right. I was jumping the gun on this one as we are still needing the upstreamed commits in `cardano-api` :/

michaelpj:
This is a case where we would benefit from https://github.com/input-output-hk/cardano-haskell-packages/issues/61

PR #69, opened 2022-10-25
dependency bumps for node 1.35.4

disassembler:
Need to include a couple more packages from base for the 2.0.0.0.1 version

Need a bug fix in consensus that prevents the intra-era hard fork from occurring


JaredCorduan (approved):
Looks good to me! The base sha is the one I expected, and the ourorboros-network sha is from https://github.com/input-output-hk/ouroboros-network/pull/4102/commits

:+1:

PR #72, opened 2022-10-26
Add static pages

andreabedini:
This PR bumps foliage to get the new generated webpages. Please test this :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,19 @@
 # Cardano Haskell package repository ("CHaP")
 
+* [Contents](https://input-output-hk.github.io/cardano-haskell-packages/contents/).
+* [Timeline](https://input-output-hk.github.io/cardano-haskell-packages/timeline/).
+
 This is a Cabal package repository ("CHaP") whose purpose is to contain all the Haskell
 packages used by the Cardano open-source project which are not on Hackage.
 
 The package repository itself is available [here](https://input-output-hk.github.io/cardano-haskell-packages).
 It is built from a [git repository](https://github.com/input-output-hk/cardano-haskell-packages) which
-contains the metadata specifying all the package versions. The package repository is built using 
+contains the metadata specifying all the package versions. The package repository is built using
 [`foliage`](https://github.com/andreabedini/foliage).
 
 If you're here because you need to add a new version of your package, you
 probably want to read the section on [adding a package from GitHub](#-from-github).
+
 
 ## What is a Cabal package repository?
 
@@ -95,7 +99,7 @@ see [below](#how-do-i-add-a-patched-versions-of-a-hackage-package-to-chap).
 
 ### Monotonically increasing timestamps
 
-When adding a package, it is important to use a timestamp (see [below](#how-to-add-a-new-package-or-package-version-to-chap)) 
+When adding a package, it is important to use a timestamp (see [below](#how-to-add-a-new-package-or-package-version-to-chap))
 that is greater than any other timestamp in the index. Indeed, cabal users rely on
 the changes to the repository index to be append-only. A non append-only
 change to the package index would change the repository index state as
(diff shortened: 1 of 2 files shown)


michaelpj:
Have we lost the landing page in the built repo?

andreabedini:
> Have we lost the landing page in the built repo?

No? foliage creates a `contents/index.html` and a `timeline/index.html`, the build script provides the `/index.html` from README.md which now has links to the two pages.

michaelpj:
Ah I see, so `foliage build` doesn't produce the full site, slightly awkward. Maybe we need a script in the repo that builds the full site and serves it?

It feels increasingly awkward having the README do double duty as the landing page for the website and the README. Not sure what to do about it.

----

These comments are all just more comments on the pages stuff, sorry :sweat_smile: 

The links are all 100% broken when opening it locally :( I can load it in a python webserver okay though. We should add instructions on how to do that.

I find the navigation quite confusing: there's no link back to the root now, so once you're in timeline or contents you're stuck there. They're two disconnected pieces, which feels odd. But I'm not sure how you'd reliably link to the root now that the user has to provide it: perhaps foliage should produce a minimal `index.html` which can be overwritten by the user, so we can assume there's *something* there to link to?

I also personally don't find "timeline" and "contents" very mnemonic at all. The timeline isn't a timeline (it can be filtered and ordered any which way), and the distinctive thing about it (in my mind) is that it's "all package versions". And "contents" isn't the contents (!) because the repository contains other versions of those packages. In my mind it's just "all packages" or "latest package versions". Even knowing what I was looking at I kept clicking between them being confused for a while.

The idea of linking to the index with a query when clicking on the breadcrumb for a package is cute, but quite non-obvious in the UI. I thought it was just broken and taking me to the wrong page for a while until I actually looked in the URL bar. I think people assume that a breadcrumb link like that will go to an actual page. Not sure what to do about it, probably fine for now.

andreabedini:
> The links are all 100% broken when opening it locally :( I can load it in a python webserver okay though. We should add instructions on how to do that.

I just don't know how to make to urls work both ways. What's the url for the root page it not "/" ? that takes you to the root of your filesystem with `file://`. Open to PR on foliage.

> I find the navigation quite confusing: there's no link back to the root now, so once you're in timeline or contents you're stuck there. They're two disconnected pieces, which feels odd. But I'm not sure how you'd reliably link to the root now that the user has to provide it: perhaps foliage should produce a minimal `index.html` which can be overwritten by the user, so we can assume there's _something_ there to link to?

Producing a minimal index.html could work. Can you suggest one?

> I also personally don't find "timeline" and "contents" very mnemonic at all. The timeline isn't a timeline (it can be filtered and ordered any which way), and the distinctive thing about it (in my mind) is that it's "all package versions". And "contents" isn't the contents (!) because the repository contains other versions of those packages. In my mind it's just "all packages" or "latest package versions". Even knowing what I was looking at I kept clicking between them being confused for a while.

I can do this renaming:
* contents -> latest package versions
* timeline -> all package versions

Note that the latter will include revisions and metadata too. Is it still ok?

> The idea of linking to the index with a query when clicking on the breadcrumb for a package is cute, but quite non-obvious in the UI. I thought it was just broken and taking me to the wrong page for a while until I actually looked in the URL bar. I think people assume that a breadcrumb link like that will go to an actual page. Not sure what to do about it, probably fine for now.

The filtering is there because it saved me making a page for all versions of a given package.

michaelpj:
> I just don't know how to make to urls work both ways. What's the url for the root page it not "/" ? that takes you to the root of your filesystem with file://. Open to PR on foliage.

Looking at what sphinx does: it just generates urls like `.../index.html` or whatever, i.e. it computes the relative path to the root page.

michaelpj:
> Producing a minimal index.html could work. Can you suggest one?

```
Cabal package repository built by foliage.

- All packages link
- All package versions link
```
i.e. extremely dumb

> The filtering is there because it saved me making a page for all versions of a given package.

Yeah, it's reasonable.

michaelpj:
Also to be clear I don't think any of the foliage changes should block this. It would be nice to have a script for building and serving the site but otherwise I think this should go in and we can iterate on it from there.

andreabedini:
@michaelpj please check this again

- Renamed the pages as discussed
- Reworked the nav bar removing the breadcrumbs
- Created simple index page
- All links are now relative to the current page

michaelpj (approved):
Looks great!

PR #75, opened 2022-10-26
Add script to check whether the newly built archive is an extension of the previous one

michaelpj:
I didn't have time to set this up in the GHA, it's changed a bit and I'm not sure what's going on right now. Maybe you can take a look @andreabedini ?

I tested this a bit and it seems to do the job. I tried:
- Adding a package with an old timestamp (BAD)
- Adding a package with a new timestamp (GOOD)
- Removing a package (BAD)
- Changing the timestamp of a package (BAD)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,16 +27,8 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    needs: check
 
     steps:
-      - uses: actions/checkout@v3
-
-      - uses: actions/cache@v3
-        with:
-          path: _cache
-          key: 1 # bump to refresh
-
       - name: Setup Nix
         uses: cachix/install-nix-action@v17
         with:
@@ -45,14 +37,40 @@ jobs:
             substituters        = https://cache.iog.io https://cache.nixos.org https://foliage.cachix.org
             trusted-public-keys = hydra.iohk.io:f/Ea+s+dFdN+3Y/G+FDgSq+a5NEWhJGzdjvKNGv0/EQ= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= foliage.cachix.org-1:kAFyYLnk8JcRURWReWZCatM9v3Rk24F5wNMpEj14Q/g=
 
+      - name: Checkout main 
+        uses: actions/checkout@v3
+        with:
+          ref: main
+             
+      - uses: actions/cache@v3
+        with:
+          path: _cache
+          key: 1 # bump to refresh
+
       - name: Unpack keys
         env:
           KEYS: ${{ secrets.KEYS }}
         run: |
-          mkdir _keys
-          echo "$KEYS" | base64 -d | tar xvz -C _keys
+          if [[ -z "$KEYS" ]]; then
+            echo "Could not access repository secret keys (PR is coming from a fork?)"
+            echo "Generating fresh keys for this run"
+            nix develop -c foliage create-keys
+          else 
+            mkdir _keys
+            echo "$KEYS" | base64 -d | tar xvz -C _keys
+          fi
 
-      - name: Build repository
+      - name: Build repository (main)
+        run: |
+          nix develop -c foliage build
+          mv _repo _repo-main
+
+      - name: Checkout tip commit
+        uses: actions/checkout@v3
+        with:
+          clean: false
+
+      - name: Build repository (tip)
         run: |
           nix develop -c foliage build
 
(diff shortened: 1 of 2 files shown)


scripts/check-archive-extension.sh
```diff
@@ -0,0 +1,74 @@
...
+	exit 1
+fi
+
+OLD_INDEX_TAR=$1
+NEW_INDEX_TAR=$2
+
+# Output looks like this:
+# -rw-r--r-- foliage/foliage  6976 2022-10-25 18:33 network-mux/0.1.0.1/network-mux.cabal
+# So fields 4 and 5 are the date and the time
+OLD_LISTING=$(tar -tvf "$OLD_INDEX_TAR" | sort -k4,5)
```
andreabedini, line 28: Sorting should not be needed because foliage always write the tarball entries in order of their timestamp.

This also simplifies things since if tarball A is a prefix of tarball B (as a list of bytes) then it's implied that all entries of tarball B have greater timestamp of all entries of tarball A. :thought_balloon: minus perhaps the edge case where the last timestamp of A is the same as the first timestamp of B :thinking:

scripts/check-archive-extension.sh
```diff
@@ -0,0 +1,74 @@
...
+function getDate (){
+  echo "$1" | tr -s ' ' | cut -f4,5 -d' ' | tr ' ' '-'
+}
+
+LAST_OLD_DATE=$(getDate "$LAST_OLD_ENTRY")
+FIRST_ACTUALLY_NEW_DATE=$(getDate "$FIRST_ACTUALLY_NEW_ENTRY")
+
+# Standard string comparison will do the job here, because
+# we're comparing dates in the format YYYY-MM-DD-HH:MM!
+if [[ $LAST_OLD_DATE > $FIRST_ACTUALLY_NEW_DATE ]]; then
```
andreabedini, line 64: I am tempted to put a `>=` here. Foliage at the moment only orders by timestamp but doesn't guarantee any order within the same timestamp. So you could add a new package with the exact same timestamp as the last one and it might get inserted just before, violating the append-only guarantee. I can make foliage give more guarantees about the ordering but I think it's a good idea to be stricter here in what we allow.

andreabedini (changes requested):
This is a good start. We are not checking the first tarball is a prefix of the second though. This check would subsumes some of the checks we do. That said, the checks you are suggesting offer good feedback, and we should keep them all anyway.

- You can assume the tarball is in timestamp order, this might simplify something
- `$LAST_OLD_DATE > $FIRST_ACTUALLY_NEW_DATE` as commented I think we should be stricter here
- Keen to add a byte-wise-comparison

scripts/check-archive-extension.sh
michaelpj: So, I tried to compare them as lists of bytes with `cmp`, but even when they were prefixes wrt entries `cmp` did not think that they were prefixes wrt bytes. Which is annoying, perhaps due to the binary structure of tar files. I figured this was easier than trying to figure that out!

(Also, sorting is cheap and robust, so why not?)

scripts/check-archive-extension.sh
michaelpj: Sounds good.

michaelpj:
> Keen to add a byte-wise-comparison

Yeah, as I said above this is what I tried first but I couldn't make it work. Here's roughly what I tried:

```
cmp -n "$(wc -c < archive-1)" archive-1 archive-2
```
i.e. compare the two up to the length of the first one. I think that should return true if the first one is a prefix of the second one, byte-wise. But generally it didn't work: there would be differences some way before the end of the first file.

michaelpj:
"A tar file is terminated by two blocks of zero bytes" that would do it...

scripts/check-archive-extension.sh
andreabedini: It would work but I'll have a go myself (tomorrow)

michaelpj:
Pushed a few changes.

michaelpj:
I figured out that of course, if the archive is predictably terminated by two blocks of size 512, I can just compare up to `length - 1024` bytes, which works nicely. Now this checks for byte-wise extension, and if that fails it goes through the other checks to try and give a better reason for why it's wrong.

michaelpj:
Have a look at the script change, but I'd like to try and get at least a basic CI check using it in this PR. I'm trying to just rebuild the base of the PR with the same keys and then compare them, that should be a reasonable check for a PR.

andreabedini:
`nix develop -c foliage create-keys` has failed because now the checkout is in `./base` and `./tip`. Perhaps it's worth using a single checkout and the cache.

Perhaps we can do something like

```
- uses: actions/checkout@v3 # we might need to add `fetch-depth: 0` or any largish number
- git checkout --detach ${{github.base_ref}}
- nix develop -c foliage build
- mv _repo _repo_base
- git checkout --detach ${{github.head_ref}}
- nix develop -c foliage build
- # compare _repo _repo_base
```

?

andreabedini:
> Okay, I think it works on PRs! I'd like to also run it on deploy, but I need to figure out what's going on there. I'm not sure it's _actually_ checking out the repo branch?

The deploy step deploys the `built-repo` artifact created by the build step, it doesn't recreate it. The `action/checkout` is because we also commit the repository to the repo branch. A bit of some wizardry there perhaps but the tl;dr; is "commit this folder _repo on top of the repo branch".

andreabedini (approved):

michaelpj:
> The deploy step deploys the built-repo artifact created by the build step, it doesn't recreate it.

Right, but the reason I want to run it on deploy is because the check that's running in `build` is currently running against a newly created repository, not the one in `repo`. _Conceivably_ they might be different, better to be safe and do it again.

> A bit of some wizardry there perhaps

Yeah, I just need to figure out a way to also get the existing version so I can check against it. I think at the moment you never do that, and you just push a new commit with the new version.

.github/workflows/CI.yml
```diff
@@ -116,17 +118,37 @@ jobs:
...
-          path: _repo
+          path: built-repo
+
+      - name: Check new index is an extension of the old index
+        run: |
+          ./src/scripts/check-archive-extension.sh repo/01-index.tar built-repo/01-index.tar
+
+      - name: Check new index is an extension of the old index
+        run: |
+          ./src/scripts/check-archive-extension.sh repo/01-index.tar built-repo/01-index.tar
```
andreabedini: Twice for good measure? :joy:

andreabedini:
It looks very clean now! Awesome work.

PR #76, opened 2022-10-26
Rework the discussion of testing CHaP

michaelpj:
Include @andreabedini's setup for just testing a built repository conveniently.


andreabedini:
I left a comment, perhaps in the wrong spot so I'll copy it here
> Should we start encouraging users to use a single index state for hackage and chap? We can leave in the comments and example with different timestamps but I think longer term it's easier to keep them in sync.

andreabedini (approved):

michaelpj:
I added some discussion of multiple index state stanzas while I was here.

PR #77, opened 2022-10-28
Add a separate build workflow step

### LINKED ISSUE #13: Consider running `build` on PRs
I think at the moment it only runs on `main`. But it's probably useful (if expensive) to run on PRs, probably some kinds of malformed metadata will only be caught by actually trying to build the repo?

michaelpj:
Fixes #13.

This adds a build step to the workflow. The built repository is uploaded as an artifact, which can then be downloaded by the `deploy` job if that runs (i.e. only on master). That gets us two things:
1. The built repository is available as a build artifact, which is possibly useful for debugging.
2. It gives us a place to put checks that need a built repository, like the archive extension check in #75.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,23 +27,7 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     needs: check
-
-    concurrency:
-      group: "pages"
-      cancel-in-progress: true
-
-    # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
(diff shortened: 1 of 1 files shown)


michaelpj:
(why are all the actions duplicated? I guess it's because it triggers on both `push` and `pull_request` :thinking: )

Okay, that seemed to work, but I guess I can't test the deploy step without merging :grimacing:

michaelpj:
I'm going to try it, I can always revert if it doesn't work.

michaelpj:
Please do take a look and leave comments Andrea, I just wanted to actually make sure it worked!

michaelpj:
Which it did, hooray.

PR #78, opened 2022-10-31
Added ekg-forward-0.3.0

coot:
From https://github.com/input-output-hk/ekg-forward at 810da2b55228a0c8a63d590c098e4c16d5c483c8

### DIFF
--- a/_sources/ekg-forward/0.3.0/meta.toml
+++ b/_sources/ekg-forward/0.3.0/meta.toml
@@ -1,0 +1,2 @@
+timestamp = 2022-10-31T17:20:22Z
+github = { repo = "input-output-hk/ekg-forward", rev = "810da2b55228a0c8a63d590c098e4c16d5c483c8" }


bolt12 (approved):

michaelpj:
Does this build with _released_ versions of ouroboros-network?

michaelpj:
I can't tell because `ekg-forward` isn't using CHaP yet.

coot:
@michaelpj I think it won't, but it will as soon as I will release `ouroboros-network` from input-output-hk/ouroboros-network#4120.  Let's wait until I'll do that.

michaelpj:
Yes, please do that first.

andreabedini:
This might be frustrating for you @coot but we can see the check that @michaelpj wrote working doing its job here.

When you opened the pull request you used the timestamp 2022-10-31T17:20:22Z which was ok at that time but since then other packages got added to the index which a timestamp later than the one you originally provided. Doing a git rebase fixed up the git history but the timestamp in the meta file is still too old. You will need to use a new timestamp for the check to pass.

I understand it's a bit confusing but we haven't found a better workflow for the moment.

michaelpj:
Hmm, I need to tweak the workflow, the helpful message didn't get printed because the script failed :D

PR #84, opened 2022-11-07
Release `ouroboros-network` packages for `cardano-node-1.35.5`

coot:
- Added network-mux-0.2.0.0
- Added ouroboros-network-0.2.0.0
- Added ouroboros-network-framework-0.2.0.0
- Added ouroboros-consensus-0.1.0.2

### DIFF
--- a/_sources/network-mux/0.2.0.0/meta.toml
+++ b/_sources/network-mux/0.2.0.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2022-11-10T09:25:21Z
+github = { repo = "input-output-hk/ouroboros-network", rev = "e9330845f1df5348d04bc6eb50c8ad62530589b2" }
+subdir = 'network-mux'
--- a/_sources/ouroboros-consensus/0.1.0.2/meta.toml
+++ b/_sources/ouroboros-consensus/0.1.0.2/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2022-11-10T09:25:21Z
+github = { repo = "input-output-hk/ouroboros-network", rev = "e9330845f1df5348d04bc6eb50c8ad62530589b2" }
+subdir = 'ouroboros-consensus'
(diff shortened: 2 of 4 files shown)


coot:
I will need to update this PR once input-output-hk/ouroboros-network#4120 is merged.

dnadales (approved):

michaelpj:
Does this all build with the version of the ledger used in 1.35.4? I'm assuming it does?

coot:
It does build with `cardano-ledger` which is in `release/cardano-node-1.35.x` branch. But there seems to be a problem: in the `release/cardano-node-1.35.x` branch ledger comes from `52da70e5a0472cd4433876289f1aebaa0c6e5c85`, and in CHaP ledger packages are from `c7c63dabdb215ebdaed8b63274965966f2bf408f`.  So I don't actually know if it builds or not.

coot:
It builds because `cardano-node` builds, that's the evidence I have.

Ugh, once we have CHaP in `ouroboros-network` it will be much easier to reason what is actually used.

michaelpj:
Well, that's going to be a problem if you try to get the `node@release/1.35` to build with these changes. So you probably want the ledger version you are using in `network@release/cardano-node-1.35.x` to be the same as the one in CHaP.

michaelpj:
> It builds because cardano-node builds, that's the evidence I have.

i.e. `node@release/1.35` builds where the only change is the inclusion of these packages as s-r-ps? or did you also change the ledger?

coot:
No ledger changes in the node; I want also to include `ouroboros-network` from `CHaP` for `node-1.35.5`.

michaelpj:
Okay, if indeed the node builds with these packages and no other changes then they must build with the corresponding version of the ledger, so it's okay. Just... tricky. Might be worth changing the ledger pin in `network@release/cardano-node-1.3.5.x` to match the node just to be on the safe side.

coot:
I agree, also the pin of `plutus` in `network@release/cardano-node-1.35.x` is wrong, `node@release/1.35` is using `plutus@1.0.0.1`.

coot:
I will not update the pins in `network@release/cardano-node-1.35.x`, it is a well without a bottom: the `cardano-base` packages are wrong as well, and it continues ... I'll just relay on the fact that I have a branch on top of `node@release/1.35` which compiles.

michaelpj:
Okay, do whatever you need to do, so long as you're confident it works :+1:

coot:
I have a proof using `s-r-p`'s, that should be enough.

PR #91, opened 2022-11-11
Echo message before trying the script that might fail

michaelpj:
So that the text gets shown even if the command fails! I'd prefer to have it at the bottom, but this will do.


andreabedini (approved):
👍

PR #93, opened 2022-11-14
Add CODEOWNERS for cardano-node and packages therein

Jimbo4350:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,37 +1,43 @@
 # cardano-node
-# No team?
+_sources/cardano-api             @input-output-hk/release
+_sources/cardano-cli             @input-output-hk/release
+_sources/cardano-client-demo     @input-output-hk/release
+_sources/cardano-node            @input-output-hk/release
+_sources/cardano-submit-api      @input-output-hk/release
+_sources/cardano-testnet         @input-output-hk/release
+
 
 # cardano-ledger
 
-_sources/byron-spec-chain        @input-output-hk/cardano-ledger 
-_sources/byron-spec-ledger       @input-output-hk/cardano-ledger 
-_sources/cardano-crypto-test     @input-output-hk/cardano-ledger 
-_sources/cardano-crypto-wrapper  @input-output-hk/cardano-ledger 
-_sources/cardano-data            @input-output-hk/cardano-ledger 
-_sources/cardano-ledger-core     @input-output-hk/cardano-ledger 
-_sources/cardano-ledger-pretty   @input-output-hk/cardano-ledger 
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -77,4 +83,4 @@ _sources/flat     @input-output-hk/plutus-core
 _sources/int-cast @input-output-hk/plutus-core
 _sources/moo      @input-output-hk/cardano-ledger
 # Need a node team
-# _sources/ekg-json   
+# _sources/ekg-json
```
michaelpj, line 86: here also

CODEOWNERS
```diff
@@ -1,37 +1,43 @@
 # cardano-node
-# No team?
+_sources/cardano-api             @input-output-hk/release
+_sources/cardano-cli             @input-output-hk/release
+_sources/cardano-client-demo     @input-output-hk/release
+_sources/cardano-node            @input-output-hk/release
+_sources/cardano-submit-api      @input-output-hk/release
+_sources/cardano-testnet         @input-output-hk/release
```
michaelpj, line 7: My understanding is that the release team is going to manage ledger and network also? So should we add them to those packages too?

michaelpj (approved):

michaelpj:
Added the team to the repo.

CODEOWNERS
Jimbo4350: @kosyrevSerge this should fall under your team

CODEOWNERS
Jimbo4350: You'll need to ask @disassembler

CODEOWNERS
deepfire: @Jimbo4350, I can't figure out where the `ekg-json` is coming from.  It's not used directly by any node package.  It's also not hosted by `input-output-hk`.

From my point of view, I have no idea why we need it.

Anyone?

michaelpj: We have issues for all patched packages: https://github.com/input-output-hk/cardano-haskell-packages/issues/56

You can see where it comes from here: https://github.com/input-output-hk/cardano-haskell-packages/blob/main/_sources/ekg-json/0.1.0.7.0.0.0.0.1/meta.toml#L2

We just took it from the `cardano-node` `cabal.project`, I believe it's a transitive dependency.

andreabedini: ekg-json is a dependency of ekg if that's what you are asking.

michaelpj:
Okay, we can leave the other things to resolve later.

CODEOWNERS
michaelpj: I think it's indirect via `cardano-tracer`?

PR #94, opened 2022-11-14
Yogeshs/fix awk issue

yogeshsajanikar:
Remove '-e' frrom the `awk` argument. It is not supported by few awk variants.


andreabedini (approved):
:+1:

PR #107, opened 2022-11-30
Reduce devShell closure size

andreabedini:
These changes reduce the closure size of devShell.default from 576.3M to 192.2M

* use the same nixpkgs from foliage, sharing is caring
* switch from `mkShell` to `mkShellNoCC` because we don't need a compiler

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,6 +16,21 @@
         "type": "github"
       }
     },
+    "blank": {
+      "locked": {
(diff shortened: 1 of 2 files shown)


andreabedini:
Now down to 185.3M

andreabedini:
133.7M !

michaelpj (approved):
Nice.

PR #109, opened 2022-12-12
Add iohk-monitoring-framework packages required by cardano-node.

erikd:


michaelpj:
It would be nice if there was a way to get a "rebase branch" button like there is for "update branch" if you require FF merges :thinking:

angerman:
Merging this now to unblock us.

PR #111, opened 2022-12-13
Add support for macos

newhoggy:


angerman:
The only issue I have with this, is that it requires the existence of `gtar` and `gdate`, which means it won't work for me :-/

newhoggy:
It doesn't require the existence of `gtar` and `gdate`.  If there is a `tar` or `date` in the `PATH` that isn't the default version of those, then it will prefer that.

angerman:
```
angerman@Moritzs-MacBook-Air> which tar
/usr/bin/tar
```

```
angerman@Moritzs-MacBook-Air> which gtar
gtar not found
```

but ultimately I'll figure out a way around.

michaelpj:
I'm okay with this if it's the generally approved way of doing this (surely we can't be the first people with this problem), but it needs a comment.

newhoggy:
I moved the BSD date/tar handling to a separate script and added a comment.

michaelpj (approved):

PR #123, opened 2023-02-10
Add a builder for packages using Nix, set cardano-node as a smoke test

michaelpj:
This adds the infrastructure to build packages from CHaP with Nix in our CI. There is a decent amount of ugliness but it works.

At the moment it doesn't try to be particularly clever: you can build a single package-version, or a bunch of them (better for parallelism!), and it's up to you to tell it what to build. But that's relatively easy to do externally once the nix stuff is in place.

@andreabedini this is my take on how we could do https://github.com/input-output-hk/cardano-haskell-packages/issues/61

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,8 +19,7 @@
   };
 
   outputs = { self, nixpkgs, foliage, flake-utils, haskell-nix, ... }:
-    let ghcVersions = [ "ghc8107" "ghc924" ];
-    in flake-utils.lib.eachDefaultSystem
+    flake-utils.lib.eachDefaultSystem
       (system:
         let overlays = [ haskell-nix.overlay ];
             pkgs = import nixpkgs { inherit system overlays; inherit (haskell-nix) config; };
--- a/flake.nix
+++ b/flake.nix
@@ -2,16 +2,65 @@
   description = "Metadata for Cardano's Haskell package repository";
 
   inputs = {
-    nixpkgs.follows = "foliage/nixpkgs";
-    foliage.url = "github:andreabedini/foliage";
+    nixpkgs.follows = "haskell-nix/nixpkgs";
+    foliage = {
+      url = "github:andreabedini/foliage";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
     flake-utils.url = "github:numtide/flake-utils";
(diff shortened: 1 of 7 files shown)


michaelpj:
I added a thing to build all the newly added packages also. I've added a commit with a fake new package to test that, we should not merge while that commit is in here.

michaelpj:
Seemed to work!

flake.nix
```diff
@@ -2,16 +2,64 @@
...
+
+            package-project = { built-repo, compiler-nix-name, pkgname, version }:
+              pkgs.haskell-nix.cabalProject' {
+                src = ./.;
+                inherit compiler-nix-name;
+                inputMap = { "https://input-output-hk.github.io/cardano-haskell-packages" = built-repo; };
+                # Note: tests and benchmarks are _not_ enabled, but exectuables are
+                cabalProjectLocal = ''
+                  extra-packages: ${pkgname}
+                  constraints: ${pkgname} == ${version}
```
andreabedini, line 37: I discovered you can use a package-id with extra-packages, avoiding the constraint.

andreabedini:
Some ideas to simplify this; basically by avoiding faking flake parameters
1. Can we list the packages inside the flake rather than outside?
2. We could be explicit about the compilers we support (only two) and create flake outputs for each compiler.

michaelpj:
> Some ideas to simplify this; basically by avoiding faking flake parameters

I think my feeling was that once we're on the hook for passing parameters to get the repository in, it's not so bad to get the caller to specify the other stuff. And it's nicer in some ways to drive the other stuff from GHA, e.g. we can do a GHA matrix over GHCs; or we can get the parallelism from Nix, if we have enough builders.

> Can we list the packages inside the flake rather than outside?

We could do this for the smoke test, and for building all the packages. I'm not sure we can do it for building only the _new_ packages, but arguably that's a stop-gap until it's feasible to build all the packages.

> We could be explicit about the compilers we support (only two) and create flake outputs for each compiler.

Yeah, that would be fine. Again, the issue is we can't have "proper" outputs so long as we're feeding the repository in.

-----

Also, any comments on how it compares to your builder approaches?

michaelpj:
Andrea's version is nicer!

PR #127, opened 2023-02-10
Only trigger the CI workflow on push if the push is to master

michaelpj:
This seems to be the way to avoid getting double workflow runs on PRs
while still getting the other behaviour we want.


andreabedini:
But it's called `main` :) I was wondering how come ouroboros-consensus-diffusion was not showing up :)

PR #140, opened 2023-02-22
plutus-core-1.1.1.0: Revise to allow for newer cardano-crypto-class

shlevy:


shlevy:
I've not yet tested this. I got the cabal file from the current git rev on chap and just changed line 287

shlevy:
To be clear, I've tested this version of plutus-core against a newer crypto-class, I have not tested this change to CHaP

zliu41 (approved):

lehins (approved):

PR #158, opened 2023-03-15
Add a script to add revisions for a package version

michaelpj:
Not very clever, relies on the existence of an already built repository.
Did a bit of script refactoring while I was there, some of which isn't
now needed, but I guess I did it already.

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -3,10 +3,7 @@
 set -o errexit
 set -o pipefail
 
-# Use gnu-tar and gnu-date regardless of whether the OS is Linux
-# or BSD-based.  The correct command will be assigned to TAR and DATE
-# variables.
-source "$(dirname "$(which "$0")")/use-gnu-tar.sh"
+SCRIPT_DIR=$(dirname "$(which "$0")")
 
 function usage {
   echo "Usage $(basename "$0") [-r REVISION] [-v VERSION] REPO_URL REV [SUBDIRS...]"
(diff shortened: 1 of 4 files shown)


andreabedini:
LGTM, I am just a bit sad that lots of logic ends up being duplicated in fogliage. Maybe I should add some "intermediary" commands in foliage. Something like 
```
$ foliage sdist --extract-cabal-file github:input-output-hk/plutus/b94d0e001c8f7350b5120b20cbc9d9021d377a8a
Fetching https://github.com/input-output-hk/plutus at b94d0e001c8f7350b5120b20cbc9d9021d377a8a
Created plutus-core-1.3.0.0.tar.gz
Created plutus-core-1.3.0.0.cabal
Created plutus-tx-plugin-1.3.0.0.tar.gz
Created  plutus-tx-plugin-1.3.0.0.cabal
Created plutus-tx-1.3.0.0.tar.gz
Created plutus-tx-1.3.0.0.cabal
Created plutus-ledger-api-1.3.0.0.tar.gz
Created plutus-ledger-api-1.3.0.0.cabal
```
This would let us reuse the Haskell logic, rather than reimplement it in bash.
:thinking:

andreabedini:
Note that this is different from the `folaige sdist` command you were thinking of in andreabedini/foliage#43. You were thinking of getting a sdist out of the published repo, while I am thinking of splitting out one intermediate step of building a repo, which is producing the sdist.

michaelpj:
I was actually thinking of exactly what you said: run a bit of the build and just spit out a specific intermediary artifact instead of the final repo.

andreabedini:
Glad we are finally on the same page the 😅

PR #154, opened 2023-03-10
Add some of plutus-apps packages

andreabedini:
This adds

* plutus-ledger-1.2.0.0
* plutus-script-utils-1.2.0.0
* freer-extras-1.2.0.0

From the commit hash https://github.com/input-output-hk/plutus-apps/commit/c4f4dc5fedd5b1804781abb7db0fb5a553e24ecb. Note this is not identical to the [1.2.0.0 release](https://github.com/input-output-hk/plutus-apps/releases/tag/v1.2.0), since unfortunately that release refers to `quickcheck-contractmodel` with the old name `contractmodel`. The only difference between the two commits is in plutus-ledger cabal file, which changes its dependency from contract-model to quickcheck-contractmodel.

```
diff --git a/plutus-ledger/plutus-ledger.cabal b/plutus-ledger/plutus-ledger.cabal
index b2426cdfc..3e9c4e46f 100644
--- a/plutus-ledger/plutus-ledger.cabal
+++ b/plutus-ledger/plutus-ledger.cabal
@@ -140,13 +140,12 @@ library
   -- defined there.
   build-depends:
     , aeson
-    , base               >=4.9  && <5
+    , base                      >=4.9  && <5
     , base16-bytestring
     , bytestring
     , cborg
     , containers
-    , contractmodel
-    , cryptonite         >=0.25
+    , cryptonite                >=0.25
     , flat
     , hashable
     , http-api-data
@@ -155,6 +154,7 @@ library
     , mtl
     , newtype-generics
     , prettyprinter
+    , quickcheck-contractmodel
     , scientific
     , serialise
     , servant
```

These three packages seem to compile with the following configuration (one day we will fix this!).

```
allow-newer:
  -- cardano-ledger packages need aeson >2, the following packages have a
  -- too restictive upper bounds on aeson, so we relax them here. The hackage
  -- trustees can make a revision to these packages cabal file to solve the
  -- issue permanently.
  , ekg:aeson

constraints:
  , cardano-api==1.35.4
  , cardano-ledger-alonzo==0.1.0.0
  , cardano-ledger-babbage==0.1.0.0
  , cardano-ledger-byron==0.1.0.0
  , cardano-ledger-core==0.1.0.0
  , cardano-ledger-shelley==0.1.0.0
  , cardano-ledger-shelley-ma==0.1.0.0

  -- cardano-prelude-0.1.0.0 needs
  , protolude <0.3.1

  -- cardano-ledger-byron-0.1.0.0 needs
[...]

PR #160, opened 2023-03-20
Add more smoke-tests packages up to cardano-node

michaelpj:
This extends the smoke-test package builder to build the newest versions of:
- `plutus-ledger-api`
- `cardano-ledger-api`
- `ouroboros-network`
- `ouroboros-consensus-cardano`
- `cardano-api`
- `cardano-node`

To make all this work I had to do a _lot_ of revisions. I took the hard but hopefully thorough route of just keeping trying to build things and then revising whatever the problem with the particular package that failed to build was. That meant a lot of"add an upper bound to foo-X -> cabal picks foo-Y -> add an upper bound to foo-Y -> ..., but I got there eventually.

The revisions are obviously very hard to review at the moment, so I guess you'll just have to trust me. If anyone wants to spot-check, I tried to list in the commit message what the revision was.

There is only one bit of global config I added: `allow-newer: ekg:json`. `cardano-node` needs it, upstream seems dead, we should probably just patch it in CHaP so we can bin that.

Note that this means that in future the CI will block new releases of these packages unless they build cleanly!

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,6 +116,7 @@ jobs:
         uses: nixbuild/nixbuild-action@v16
         with:
           nixbuild_token: ${{secrets.NIXBUILD_TOKEN}}
+          generate_summary_for: 'job'
 
       - uses: actions/checkout@v3
 
--- a/_sources/byron-spec-chain/0.1.1.1/meta.toml
+++ b/_sources/byron-spec-chain/0.1.1.1/meta.toml
@@ -1,3 +1,7 @@
 timestamp = 2022-11-11T00:24:50Z
 github = { repo = "input-output-hk/cardano-ledger", rev = "2fac95ef338ff9622543d08fe16d0fa6716d0019" }
 subdir = 'eras/byron/chain/executable-spec'
+
+[[revisions]]
+number = 1
+timestamp = 2023-03-21T14:36:24Z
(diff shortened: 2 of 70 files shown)


michaelpj:
(The build was fast because I've been experimenting by building on nixbuild.net, so it was already cached)

andreabedini:
> There is only one bit of global config I added: `allow-newer: ekg:json`. `cardano-node` needs it, upstream seems dead, we should probably just patch it in CHaP so we can bin that.

Unfortunately the situation is complex, see [my flawed attempt](https://github.com/haskell-infra/hackage-trustees/issues/358) of making a revision for ekg. tl;dr: This is a consequence of us forking ekg-json. We fixed ekg-json but ekg still need relaxing its bounds. Maybe it makes sense to relax the bounds anyway? ekg *does* work with `aeson>=2` *as long as* ekg-json works with `aeson>=2`. I might reopen that issue and check what the trustees think about this.

andreabedini:
@michaelpj I am not sure whether you have already done this or not but I am going to suggest it anyway for the sake of institutional knowledge.

When we make a revision for pkg-a, we need to make sure that our repos where pkg-a is imported keep working. I.e. we could have accidentally put bounds that are too restrictive. This is what happened in #154 and the fix is still WIP (see https://github.com/input-output-hk/plutus-apps/pull/1023, the typo in freer-extras.cabal is still in CHaP :facepalm:).

michaelpj:
@andreabedini one data point: I was able to build `cardano-node` master with this, including after deleting most of the `constraints` stanza. I'll try and check plutus-apps too.

michaelpj:
I'm going to merge this as I think it's an improvement and I will iterate from here if there are problems.

PR #162, opened 2023-03-22
README updates

michaelpj:


michaelpj:
Hmm, that check will only work if you PR from the repo, how annoying.

PR #167, opened 2023-03-23
Publish 1.4.0.0 for plutus core packages

zliu41:

### DIFF
--- a/_sources/plutus-core/1.4.0.0/meta.toml
+++ b/_sources/plutus-core/1.4.0.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2023-03-25T12:00:00Z
+github = { repo = "input-output-hk/plutus", rev = "411cf29bfbffbab6a4e6a31ef783cd52362affdf" }
+subdir = 'plutus-core'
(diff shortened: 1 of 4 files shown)


effectfully:
> Alright, this is a real failure from the new CI, hooray. We need to deal with https://github.com/obsidiansystems/dependent-sum/issues/72 now in a way that works for downstream without requiring config.

`deriveGEq` was removed and I guess we could just drop `deriveGCompare` . It would be very annoying for some of our users, since we added it because there was a request, not due to us needing it, but if it's a serious trouble, we could make that sacrifice. Alternatively, we could implement the `GCompare` instance ourselves, but that's a ton of boilerplate, but maybe we could think on a clever approach.

michaelpj:
The alternative is just to add a spurious dep on `dependent-sum` so we can enforce the correct lower bound.

andreabedini:
I want to understand why we are not picking a more recent `dependent-sum-` here. On my machine, I removed the constraint from cabal.project and cabal picked `dependent-sum-0.7.1.0` :thinking:

PR #169, opened 2023-03-24
New revisions of ouroboros-consensus-shelley and cardano-ledger-shelley

coot:
* `ouroboros-consensus-shelley-0.1.0.1` is compatible with
  `ouroboros-network < 0.4`.
* Relax upper bound of `cardano-data` in `cardano-ledger-shelley` to
  allow all minor versions `cardano-data-0.1.x`.

### DIFF
--- a/_sources/cardano-ledger-shelley/0.1.0.0/revisions/3.cabal
+++ b/_sources/cardano-ledger-shelley/0.1.0.0/revisions/3.cabal
@@ -1,0 +1,119 @@
+cabal-version: 2.2
+
+name:                cardano-ledger-shelley
+version:             0.1.0.0
+description:         Shelley Ledger Executable Model
+license:             Apache-2.0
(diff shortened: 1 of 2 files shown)


michaelpj (approved):
LGTM as long as the checks pass.

michaelpj:
(These are relaxing bounds that I put there with revisions and are clearly too stringent)

coot:
Would this be caught by `build-new-packages` which we ignore for these versions?

coot:
I double checked that `cardano-data-0.1.0.1` was used by `cardano-node-1.35.6`.

michaelpj:
`build-packages` will make sure that the released `cardano-node`s continue to build, which is what's important.

michaelpj:
`build-new-packages` doesn't currently notice that it should build revised packages, unfortunately.

PR #179, opened 2023-03-27
`fs-api-0.1.0.0`, `fs-sim-0.1.0.0`

jorisdral:
Part of https://github.com/input-output-hk/ouroboros-network/issues/4082


andreabedini:
The failure is because CI has a fixed version of hackage.nix, which is old enought to not include `unix-bytestring-0.4.0` (released about 10 days ago). @michaelpj maybe we could add `--update-input hackage-nix` to always build with the newest hackage?

michaelpj:
Mmmm, I somewhat want to do that deliberately? It's going to force a *lot* of rebuilding in general. So I think I'm happy for us to respond to this with "guess we need to update `hackage.nix`", go do that (dealing with any problems), and then come back to this.

dnadales (approved):

michaelpj:
Indeed, bumping Hackage in this case meant dealing with the fallout of `unix-bytestring-0.4.0.0`, which meant more revisions to `ouroboros-consensus`, so I'm glad that didn't land on this PR :)

PR #181, opened 2023-03-29
Publish io-sim v0.5.0.0

dnadales:


michaelpj:
Looks like this breaks building latest ouroboros-network. Probably the broken ouroboros-network packages need upper bounds on the relevant io-sim package, which you can fix by adding a revision.

coot (approved):

coot:
Right, and also `typed-protocols` seemed to be broken.

coot:
It seems we are building `io-sim-0.5.0.0` with an incompatible version of `io-classes`.  @dnadales we should have used `io-sim ^>= 0.5` in `io-classes.cabal` file.

coot:
I passed a revision to `typed-protocols` which you will anyway need when updating `ouroboros-network` packages.

dnadales:
> It seems we are building `io-sim-0.5.0.0` with an incompatible version of `io-classes`. @dnadales we should have used `io-sim ^>= 0.5` in `io-classes.cabal` file.

Ouch. So we should probably update all cabal files to use the caret operator instead, right?

michaelpj:
You should generally pin at least the major version of any IOG libraries you depend on - they tend to break often! Caret operator is a good way to do taht.

michaelpj (approved):

PR #182, opened 2023-03-30
Automate bumping hackage.nix

### LINKED ISSUE #166: Add a workflow to update the flake inputs occasionally
Use https://github.com/DeterminateSystems/update-flake-lock .

In particular, we want to update our `hackage.nix` pin relatively frequently so we're generally building against a new-ish Hackage. The others we can probably update manually.

andreabedini:
Closes #166

We defintely want to run tests here so we need to set a token as in the `DeterminateSystems/update-flake-lock` [README](https://github.com/DeterminateSystems/update-flake-lock#with-a-personal-authentication-token).

### DIFF
--- a/.github/workflows/update-flake-lock.yml
+++ b/.github/workflows/update-flake-lock.yml
@@ -1,0 +1,24 @@
+name: Bump hackage.nix
+
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '0 0 * * *' # runs daily at 00:00
+
+jobs:
+  lockfile:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v1
+      - name: Update flake.lock
+        uses: DeterminateSystems/update-flake-lock@v18
+        with:
+          inputs: hackage-nix
+          pr-labels: |
+            dependencies
+            automated
+      - name: Print PR number
+        run: echo Opened pull request ${{ steps.update.outputs.pull-request-number }}.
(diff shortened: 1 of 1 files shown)


.github/workflows/update-flake-lock.yml
```diff
@@ -0,0 +1,24 @@
+name: Bump hackage.nix
+
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '0 0 * * *' # runs daily at 00:00
```
michaelpj, line 6: weekly?

.github/workflows/update-flake-lock.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v1
+      - name: Update flake.lock
+        uses: DeterminateSystems/update-flake-lock@v18
+        with:
+          inputs: hackage-nix
```
michaelpj, line 19: also `CHaP`? Since there is an input, we might as well update it occasionally! Even thought we generally override it.

michaelpj:
Do you want to make a token? Otherwise I can do it.

andreabedini:
I tried to make a PAT only valid for `input-output-hk/cardano-haskell-packages` but I cannot do that.

michaelpj:
I don't know what to do from here then :confused:

andreabedini:
Let's go ahead anyway, we have `ci.zw3rk.com` hooked up and that will build all PRs anyway.

michaelpj:
> Let's go ahead anyway, we have ci.zw3rk.com hooked up and that will build all PRs anyway.

And conveniently, that will do what we want here since it'll use the CHaP from the flake input, which is exactly what'll be updated in this PR :joy:

PR #183, opened 2023-03-30
Deal with ekg, ekg-json, and libsystemd-journal

michaelpj:
These are the last remaining `allow-newer`s in `cardano-node`'s `cabal.project`.

I did the following:
- Added a patched version of `ekg` based on upstream (so it's clear it's going from a safe repository)
    - Revised it to relax bounds
- Revised our patched `ekg-json` to relax some bounds
- Added a patched version of `libsystemd-journal` based on upstream
    - Revised it to relax bounds

Tested that I can build `cardano-node` on 9.6 with this repository and it now needs no config.


michaelpj:
This also lets us get rid of the only bit of global config we had in the CI builder: `allow-newer: ekg:json`.

PR #184, opened 2023-03-30
typed-protocols: relax io-sim version bound

coot:


coot:
This would only be a problem if one has too old `index-state`, but then moving it forward will bring `typed-protocols-0.1.0.3`.

We do have constraints on `io-classes`, so we're fine.

michaelpj:
Not necessarily. If cabal can choose `io-classes-0.4` and `typed-protocols-0.1.0.2` OR `io-classes-0.5` and `typed-protocols-0.1.0.3`, then there's no law saying it has to pick the latter. I think it _probably_ will, since it usually tries to pick later versions, but I have been surprised by it doing things like that in the past. But if you will have a bound on `io-classes` downstream then that's okay.

michaelpj:
Yeah, I'm not really sure what "good practice" looks like here yet :shrug:

PR #188, opened 2023-04-03
Add `marlowe`, `marlowe-spec-test`

jhbertra:
Add packages from https://github.com/input-output-hk/marlowe


michaelpj:
I'm guessing your packages are unbuildable with GHC 9.2. Can you add some lines for your packages similar to the existing ones here: https://github.com/input-output-hk/cardano-haskell-packages/blob/main/flake.nix#L70

jhbertra:
We actually don't need to publish that package, so I'm going to try an simply remove it.

PR #189, opened 2023-04-03
Patch `servant-server`, `vector-algorithms`, `hedgehog`

jhbertra:
Patches for three hackage packages to allow newer for `resourcet` and `primitive`.


shlevy (approved):
Assuming the base versions match the upstream git tags, LGTM

jhbertra:
> Assuming the base versions match the upstream git tags, LGTM

They all do, with the exception of `vector-algorithms` which @erikd did not tag. In that case, the SHA points to the commit in which the cabal version number was changed to 0.9.0.1.

jhbertra:
> From the README:
> 
> > IMPORTANT: if you release a patched package to CHaP, make sure to open an issue about it so we can keep track of which patched packages we have. Ideally, include the conditions under which we can deprecate it, e.g. "can deprecate either when it's fixed upstream or when package X removes their dependency on it".

Done

michaelpj:
Thanks for documenting!

PR #197, opened 2023-04-05
Bump haskell.nix

michaelpj:
This pulls in Andrea's recent fix which should stop the CI from doing so
many spurious rebuilds.


michaelpj:
Built the repo locally and got a cache hit, hooray!

PR #200, opened 2023-04-06
Rework builder to work with new haskell.nix

michaelpj:
Now haskell.nix fetches the sdists from the actual repository URL, not the flake input. But for new packages they won't be _in_ the remote version of `cardano-haskell-packages`, so it will fail. Instead we use the flake input as the repo directly.

Contains a fake package publish just to test the CI. Do not merge with that in.

Do not merge until https://github.com/input-output-hk/haskell.nix/pull/1910 is merged also.


michaelpj:
Okay, I take it back, it doesn't work.

michaelpj:
A pleasingly small diff in the end.

PR #202, opened 2023-04-11
Add a revision for `libsystemd-journal` to allow newer `unix-bytestring`

lehins:
I was able to successfully build `cardano-node` with newest `unix-bytestring-0.4`. Only bounds that are updated in this revision are:
```diff
-         unix-bytestring >=0.3.2 && <0.4,
+         unix-bytestring >=0.3.2 && <0.5,
```

### DIFF
--- a/_sources/libsystemd-journal/1.4.5.0.0.0.0.1/meta.toml
+++ b/_sources/libsystemd-journal/1.4.5.0.0.0.0.1/meta.toml
@@ -5,3 +5,7 @@ force-version = true
 [[revisions]]
 number = 1
 timestamp = 2023-03-31T11:59:45Z
+
+[[revisions]]
+number = 2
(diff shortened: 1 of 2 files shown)


michaelpj:
Hmm, that's surprising, since didn't we just add a bunch of upper bounds on this for other packages of ours? Anyway, the CI is happy I guess.

lehins:
@michaelpj I suspect [`unix-bytestring-0.4`](https://hackage.haskell.org/package/unix-bytestring)  simply wasn't noticed in the previous revision. It turns out `fs-api` requires it now, so we must use that new version of `unix-bytestring` which was released less than a month ago: https://github.com/input-output-hk/fs-sim/blob/dec2adf9721e27597e39642995895863a024d988/fs-api/fs-api.cabal#L66

PR #205, opened 2023-04-12
Add a catch-all CODEOWNERS rule for trustees

michaelpj:
At the moment we're using CODEOWNERS to manage access control. That
works for what it covers, but some stuff _isn't_ covered.

That means that anyone who is part of the teams with write access can,
e.g. push a new package, including patched Hackage packages; add new
package versions of anything that isn't owned; alter the CI; etc.

There's a fairly simple solution: just add a catch all rule for the CHaP
"trustee" team (a new team I just made so that it can be distinct from
the team which has admin on the repo).  This means that everyting will
get reviewed by _someone_.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,3 +1,7 @@
+# Anything that isn't explicitly owned by a team 
+# needs approval from the admins
+* @input-output-hk/cardano-haskell-packages-trustees
+
 # cardano-node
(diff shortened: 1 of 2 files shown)


andreabedini (approved):
:+1: asssuming the trustees can indeed be trusted :joy_cat:

PR #209, opened 2023-04-14
Relax upper bounds on the plutus-ledger-api in ledger packages

lehins:


michaelpj:
Seems like it fails.

lehins:
> Seems like it fails.

@michaelpj Yep, I am working on that

PR #211, opened 2023-04-17
Add quickcheck-contractmodel 0.1.3

berewt:

### DIFF
--- a/_sources/quickcheck-contractmodel/0.1.3.0/meta.toml
+++ b/_sources/quickcheck-contractmodel/0.1.3.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2023-04-24T15:30:00Z
+github = { repo = "input-output-hk/quickcheck-contractmodel", rev = "8d24361898201d628032f21a744a8ef2da23213f" }
+subdir = 'quickcheck-contractmodel'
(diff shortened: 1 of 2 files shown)


sjoerdvisscher:
That build error seems to suggest that a dependency of `plutus-core` (`ral`) needs an older GHC version to build? @michaelpj does that ring a bell?

michaelpj:
```
quickcheck-contractmodel> [__1] skipping: plutus-tx-1.4.0.0, plutus-tx-1.3.0.0, plutus-tx-1.2.0.0,
quickcheck-contractmodel> plutus-tx-1.1.1.0, plutus-tx-1.1.0.0 (has the same characteristics that caused
quickcheck-contractmodel> the previous version to fail: excluded by constraint '^>=1.0' from
quickcheck-contractmodel> 'quickcheck-contractmodel')
```

You're pinned to a very old version of `plutus-tx` that is before we added support for GHC 9.2. You should probably relax that. Or we can say that `quickcheck-contractmodel` isn't buildable on 9.2, but perhaps better to fix it.

berewt:
Yes, we replicated the boundaries of `plutus-apps` to ensure compatibility. There's no much we can do about it at the moment (we definitely need to find a way to upgrade them in a near future).

So, if we want to stick to ghc8.x at the moment, how should I flag the package?

michaelpj:
There's a somewhat grimy listing of things that aren't buildable here: https://github.com/input-output-hk/cardano-haskell-packages/blob/main/flake.nix#L70

PR #213, opened 2023-04-18
Add a script to update CHaP dependencies

michaelpj:


michaelpj:
A piece of the automation puzzle.

andreabedini:
I am not sure I understand the plan here. This is a script to update the bounds in the cabal files.

- Is this supposed to be run on the other repositories? 
- I guess one still needs to build the projects to see if that worked? (Notice there could be constraints in project files)
- How is this different from doing `cabal build all --allow-newer`?

andreabedini:
Gotcha, happy to go ahead and we will see that sticks :) any cache we can hook this up to the flake so people can do `nix run github:input-output-hk/cardano-haskell-packages#update-my-deps`?

michaelpj:
Okay, I added some docs, and I added it to the flake. You _can_ run it nicely that way, if you use `--update-input` and `--no-write-lock-file` then it will run in the context of the latest CHaP, which is neat.

PR #214, opened 2023-04-19
Correct the release date of qc-contractmodel

berewt:
I can't read a calendar, unfortunately.


michaelpj:
You cannot do this, it has been released already. The timestamp is fine, no need to mess with it (you _did_ use `add-from-github.sh`, right?)

berewt:
I didn't. 😖
(and I used a wrong date, but we have a workaround, so let's keep it)

michaelpj:
It's not a case of choosing whether to keep it or not: the index is immutable, we simply cannot change it now.

PR #217, opened 2023-04-20
Update cardano-prelude and cardano-prelude-test

erikd:


andreabedini:
@michaelpj what's the story with access to `nixbuild_token`? I there a way I can give the tests access to it or we need to give Erik some extra permission?

PR #218, opened 2023-04-20
Released necessary ouroboros-network-* packages for https://github.com/input-output-hk/cardano-node/pull/4854

bolt12:
From https://github.com/input-output-hk/ouroboros-network at aabf8f87f7587b5e94f109c6ab64e02d1db03c35

Still missing a release for `ouroboros-consensus` and `ouroboros-consensus-diffusion`. @Jasagredo , it should be the same revision.


coot:
@Jasagredo We also need a revision of `ouroboros-consensus-byron`. We needed `ouroboros-consensus-diffusion     ^>=0.3 || ^>=0.4` in its `*.cabal` file.

coot (approved):

Jasagredo:
@coot Or a release that removes the dependency completely :smile: It is only needed for `db-converter` which is deep legacy stuff.

PR #231, opened 2023-04-27
Revision #1 for set-algebra to use cardano-data <1.1

aniketd:
A previously exposed module and some instances were removed from `cardano-data >=1.1` in [cardano-ledger/#3371](https://github.com/input-output-hk/cardano-ledger/pull/3371).

This PR adds revisions to all previously released versions of `set-algebra` to restrict its dependency on `cardano-data` to be `<1.1`.


lehins (approved):
Perfect! Thank you.

PR #235, opened 2023-04-29
Add note about troubleshooting CI builds

angerman:
Hope this helps the next poor soul coming across this.


michaelpj:
Yep, or you can just build the repositories locally.

michaelpj:
Yes, that's true. But the keys don't really matter much, and I _think_ they don't affect the eventual builds, since those only depend on the source distributions and cabal files.

PR #249, opened 2023-05-09
Publish 1.6.0.0 for plutus core packages

bezirg:


zliu41:
@bezirg please update the timestamps

PR #260, opened 2023-05-11
Add shortcut to quickly add constraints to packages

andreabedini:
This adds a function to the derivations passthru to be able to quickly add constraints to builds.

E.g. if `.#"ghc8107/cardano-ledger-core/0.1.0.0"` is an derivation that aggregates all components of `cardano-ledger-core-0.1.0.0` then

```
.#"ghc8107/cardano-ledger-core/0.1.0.0".passthru.addConstraint "cardano-crypto-class <2.0.0.1, cardano-binary<1.5.0.1"
```
builds the same aggregates with the additional listed constraints.

Usage from the command line is a bit awkward because nix being ... nix, but we can script over it:

```
nix build $(nix eval --raw '.#"ghc8107/cardano-ledger-core/0.1.0.0"' \
    --apply 'd: (d.passthru.addConstraint "cardano-crypto-class <2.0.0.1, cardano-binary<1.5.0.1").drvPath')
```

Unfortunately there's no `nix build --apply` ([yet](https://github.com/NixOS/nix/pull/6333)), so I have to do it in two steps.

### DIFF
--- a/nix/builder.nix
+++ b/nix/builder.nix
@@ -16,6 +16,14 @@ let
     let
       package-id = "${package-name}-${package-version}";
 
+      # Global config needed to build CHaP packages should go here. Obviously
+      # this should be kept to an absolute minimum, since that means config
+      # that every downstream project needs also.
+      #
+      # No need to set index-state:
+      # - haskell.nix will automatically use the latest known one for hackage
+      # - we want the very latest state for CHaP so it includes anything from
+      #   e.g. a PR being tested
       project = (pkgs.haskell-nix.cabalProject' {
         inherit compiler-nix-name;
 
@@ -42,6 +50,19 @@ let
         }];
       });
 
+      # Wrapper around all package components
+      #
+      # The wrapper also provides shortcuts to quickly manipulate the cabal project.
+      #
+      # - addCabalProject adds arbitrary configuration to the project's cabalProjectLocal
+      # - addConstraint uses addCabalProject to add a "constraints: " stanza
+      # - allowNewer uses addCabalProject to add a "allow-newer: " stanza
+      #
+      # Note 1: We use cabalProjectLocal to be able to override cabalProject
+      # Note 2: `cabalProjectLocal` ends up being prepended to the existing one
+      # rather than appended. I think this is haskell.nix bug. If the project
+      # has already a `cabalProjectLocal` this might not give the intended
+      # result.
       aggregate = project:
         pkgs.releaseTools.aggregate
           {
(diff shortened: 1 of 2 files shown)


nix/builder.nix
```diff
@@ -51,13 +51,40 @@ let
...
-      constituents = components;
-      # pass through the plan for debugging purposes
-    } // { passthru = { inherit (project) plan-nix; }; }; 
-in build-chap-package
+      components = builtins.map
+        (c: lib.attrsets.recursiveUpdate c {
+          passthru = {
+            addConstraint = constraint:
+              let
+                project' = project.appendModule { cabalProjectLocal = "constraints: ${constraint}"; };
```
michaelpj, line 59: Why not `addCabalConfig` and drop the `constraints: ` bit, so you can just stick whatever additional config line you want in there? Sometimes you might want something else, e.g. an `allow-newer`.

nix/builder.nix
```diff
@@ -51,13 +51,40 @@ let
...
+              in
+              # this is a ugly way to find the same component again
```
michaelpj, line 61: :see_no_evil:

nix/builder.nix
```diff
@@ -51,13 +51,40 @@ let
...
-      # actually run tests: CHaP will not check that your tests pass (neither
-      # does Hackage).
-      constituents = components;
-      # pass through the plan for debugging purposes
-    } // { passthru = { inherit (project) plan-nix; }; }; 
-in build-chap-package
+      components = builtins.map
+        (c: lib.attrsets.recursiveUpdate c {
+          passthru = {
+            addConstraint = constraint:
```
michaelpj, line 57: Should we upstream this, seems kind of useful?

nix/builder.nix
```diff
@@ -51,13 +51,40 @@ let
...
+        (haskellLib.getAllComponents pkg);
+
+      aggregate = pkgs.releaseTools.aggregate
+        {
+          name = package-id;
+          # Note that this does *not* include the derivations from 'checks' which
+          # actually run tests: CHaP will not check that your tests pass (neither
+          # does Hackage).
+          constituents = components;
+          # pass through the plan for debugging purposes
```
michaelpj, line 77: in the wrong place now

michaelpj (approved):
Needs some doc, at least what you wrote in the PR description should go somewhere :)

nix/builder.nix
andreabedini: Opened https://github.com/input-output-hk/haskell.nix/issues/1951

nix/builder.nix
andreabedini: Yes, eventually :)

michaelpj (approved):

PR #264, opened 2023-05-12
Ouroborous related packages using iog-contra-tracer in place of contra-tracer.

erikd:


andreabedini:
In this intended to be an actual release or just testing? Those hashes are not tagged as a release in the repository and risk being garbage collected by github.

erikd:
> In this intended to be an actual release or just testing? Those hashes are not tagged as a release in the repository and risk being garbage collected by github.

Current `master` of the `ouroboros-network` packages are not API compatible with the packages further up the stack so I am "patching" the least release that is.

PR #275, opened 2023-05-15
Add revision for `ouroboros-network-protocols 0.5.0.1`

jasagredo:
`ouroboros-network-protocols` 0.5.0.1 is not compatible with `ouroboros-network-api` 0.3.0.0.

Also adds `ouroboros-consensus-diffusion` as a top-level package for the smoke tests.


coot:
We also need to revise `ouroboros-network-protocols-0.5.0.0`.

michaelpj (approved):
Looks like it built the latest`ouroboros-consensus-diffusion` fine with this.

michaelpj:
(I'm assuming you're going to merge this when you're happy and have added the extra revision)

jasagredo:
I think the conclusion from todays Consensus office hours was that this is not needed. Right @coot ?

coot:
Yes, this is not needed, I deprecated `5.0.0.1`.

PR #277, opened 2023-05-16
scripts/add-revision.sh: Improve error reporting

erikd:
I looked into actually retrieving the needed cabal file but that seems a little too fragile, so report the error, create the required  directory and suggest copying the needed cabal file.


angerman (approved):
LGTM, however, it's not clear to me yet how to end up in this situation? Why wouldn't the current cabal file exist?

erikd:
> LGTM, however, it's not clear to me yet how to end up in this situation? Why wouldn't the current cabal file exist?

Its the cabal file for the package we are modifying. That directory is populated by (as I understand it) a Nix program called `folliage`. If you don't have Nix, then you will not have this file,

michaelpj:
I guess this is fine. You can also just download the repository and pass that if you don't want to use foliage.

michaelpj:
@erikd `foliage` is a Haskell program. You can go `cabal build` https://github.com/andreabedini/foliage if you don't want to use Nix.

PR #279, opened 2023-05-16
network/release 2023 05 16

coot:
- Added ouroboros-network-0.7.0.1
- Added ouroboros-network-api-0.5.0.0
- Added ouroboros-network-framework-0.6.0.1
- Added ouroboros-network-protocols-0.5.0.2

### DIFF
--- a/_sources/ouroboros-network-api/0.5.0.0/meta.toml
+++ b/_sources/ouroboros-network-api/0.5.0.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2023-05-16T06:01:39Z
+github = { repo = "input-output-hk/ouroboros-network", rev = "130d7aa31a8da93e1f4bba3ecc9b937251f8aec2" }
+subdir = 'ouroboros-network-api'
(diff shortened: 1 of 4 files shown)


coot:
* [x] checked with  in

coot:

michaelpj:
> checked with build-with-chap.sh in ouroboros-network

Is there something we should try to work into the CHaP CI, or is this inherently an upstream repository thing?

coot:
Few weeks ago I made a release, but forgot to release one package (they say don't release things on Fridays, especially when one goes on vacations :grin:).  Then I found out that the consensus team has a script which can help with this, I generalised it so it can work for any repo: [build-with-chap.sh](https://github.com/input-output-hk/ouroboros-network/blob/3d01a31e8094e0d9bd921f966ecb474bbfacd0fc/scripts/build-with-chap.sh).

coot:
The idea is that it resets packages inside `ouroboros-network` to the most recently published version and builds them (:thinking: it should also run the tests).  We could have something like that in CHaP, but not for all packages just ones that come from the same repository.

michaelpj:
Okay, so we already build the newly added packages in the CHaP PR, so it seems like the only thing this can catch is a problem in an unreleased package? So I wonder if it would just be simpler to have a check for that (or maybe wrap `add-from-github` to always add all the packages you care about?).

coot:
> so it seems like the only thing this can catch is a problem in an unreleased package?

Yes, I think you're right and that was my intention.

> So I wonder if it would just be simpler to have a check for that (or maybe wrap add-from-
github to always add all the packages you care about?)

But for that we need to know which packages we care about in the given release in the first place.  We don't release everything in each release, because some of our packages are quite stable, and thus the script comes quite handy. We could probably just limit it to `cabal build --dry-run all` but I was thinking to expand it in the other direction: to run tests suites, something we probably don't want to do in `CHaP`.

PR #291, opened 2023-05-23
Release crypto with bls

lehins:
BLS work is done on the crypto side, so it is time to finally release it.


andreabedini:
```
       > [__0] rejecting: cardano-crypto-class-2.1.2.0 (conflict: pkg-config package
       > libblst-any, not found in the pkg-config database)
```

This needs a haskell.nix bump.

andreabedini (approved):
LGTM :+1: We only need to make it work :joy:

lehins:
Ok, I have no idea why it can't find `blst`. @angerman You got any ideas?

PR #293, opened 2023-05-24
Release consensus 0.7.0.0 and friends

amesgen:


jorisdral (commented):
Our previous releases used a commit as the revision, we're now using a tag. What would we prefer?

amesgen:
Ah good point, tags are technically mutable (although we don't plan on doing so). I will revert to using commits for consistency, good catch :+1:

dnadales (approved):

PR #296, opened 2023-05-25
fs-sim and fs-api

erikd:


jorisdral:
This release might cause some breakage down the line because `fs-api` now depends on `unix-2.8` or higher. I'm building a PR to make `fs-api` compatible with older versions of `unix`, so we can include that in this release

jorisdral (changes requested):
See previous comment

jorisdral:
Done. Could you use https://github.com/input-output-hk/fs-sim/commit/0930a32ed4641d93a253c83996cb229789bcd073 instead?

PR #316, opened 2023-06-01
node 8.1.1 packages

JaredCorduan:


michaelpj:
`tx-generator` seems to now depend on `plutus-scripts-bench` which doesn't exist.

`trace-dispatcher-2.0` doesn't build, some change here must have allowed it when it wasn't before.

disassembler (approved):

PR #322, opened 2023-06-05
Update README.md

angerman:

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,9 +232,18 @@ There is a convenience script for adding a revision to CHaP:
 $ ./scripts/add-revision.sh _repo PACKAGE_NAME PACKAGE_VERSION
 ```
 
-You need a [built package repository](#how-to-build-the-cabal-package-repository) in order to use the script.
-It will add a new revision and copy the _current_ cabal file in as the revised cabal file.
-You can then edit that file and commit the result.
+The full workflow looks like this:
+```bash
+git clone git@github.com:input-output-hk/cardano-haskell-packages.git && cd cardano-haskell-packages
+git checkout -b $USER/${PACKAGE_NAME}-rev
+curl -LO https://input-output-hk.github.io/cardano-haskell-packages/01-index.tar.gz
+mkdir -p _repo/index && tar -C _repo/index -xzf 01-index.tar.gz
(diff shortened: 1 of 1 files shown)


erikd (approved):

README.md
```diff
@@ -232,9 +232,18 @@ There is a convenience script for adding a revision to CHaP:
 $ ./scripts/add-revision.sh _repo PACKAGE_NAME PACKAGE_VERSION
 ```
 
-You need a [built package repository](#how-to-build-the-cabal-package-repository) in order to use the script.
-It will add a new revision and copy the _current_ cabal file in as the revised cabal file.
-You can then edit that file and commit the result.
+The full workflow looks like this:
```
michaelpj, line 235: Why not put this information in the section about getting a built package repository that was linked to previously? You just need to say "instead of building it with foliage you can download it by running ..."

Also, maybe it's just documentation philosophy, but I'm not a huge fan of having all the other steps listed out. This README lists many things you might want to do, all of which require having the repo checked out. Do you want to add "do a git clone and cd into it" to _every_ one? That would be consistent at least! But then let's decide to do that instead of having random examples that are far more detailed than the others.

angerman: Yes. This is messy. What I'm learning is that CHaP in itself has no value. It's a vehicle and most don't really have time to study CHaP in detail.

What we seemingly need to go towards is a bunch of items:

- I want to do X
  <explicitly list the steps needed>
...

I think this is mostly a psychological thing. I'd you see right away it's just something that's going to cost you 5min, you are much more inclined to do it. If it looks like it might be more involved, the incentive can be to just not do it, postpone it; or hope someone else does it. However we want everyone to feel like it's an easy and enjoyable task that can be done quickly--even without understanding CHaP, or having to understand it in detail.

michaelpj: So currently the "requirements" for using CHaP are: check out the repo; spend ~10 mins reading the README which explicitly lists how to do most tasks; use the nix shell to get foliage if you need it. Is that so bad?

If you want to have a set of how-tos that are completely explicit with every step for how to do the thing then sure, we can do that, but it'll make the README much longer and people already don't read it.

angerman:
Will close this for a different approach as discussed with @michaelpj

PR #324, opened 2023-06-05
Add some notes on debugging solver problems

michaelpj:

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,7 +429,8 @@ This will run the script at every step of the rebase on `HEAD` (i.e. the commit 
 ### Debugging solver issues
 
 Sometimes you may hit issues that are related to cabal's constraint solver making strange choices.
-For example, you are releasing `foo-X`, which depends on `bar` via `baz`, and instead of using the newest `bar-Y`, cabal inexplicably decides to build a very old verison of `bar` that doesn't work for some reason.
+For example, you are making a PR to release `foo-X`, which depends on `bar` via `baz`. 
+But when the CI builds the package, instead of using the newest `bar-Y`, cabal inexplicably decides to build a very old verison of `bar` that either a) leads to solver errors; or b) leads to compilation errors.
 
 The root cause is usually that the solver can't pick the version of `bar` that you want because it conflicts with your dependencies in some way, but it is often not obvious why.
 It's tricky to diagnose, since you'll sometimes want to rebuild the repository to add revisions while you're working.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -426,4 +426,29 @@ There are some scripts for dealing with this:
 An easy way to run `update-timestamps-and-fixup` on a multi-commit PR is to run `git rebase main --exec "./scripts/update-timestamps-and-fixup.sh HEAD"`.
 This will run the script at every step of the rebase on `HEAD` (i.e. the commit you have reached).
 
+### Debugging solver issues
+
+Sometimes you may hit issues that are related to cabal's constraint solver making strange choices.
+For example, you are releasing `foo-X`, which depends on `bar` via `baz`, and instead of using the newest `bar-Y`, cabal inexplicably decides to build a very old verison of `bar` that doesn't work for some reason.
```
andreabedini, line 432:
```suggestion
For example, you have released `foo-X`, which depends on `bar` via `baz`, and instead of using the newest `bar-Y`, cabal inexplicably decides to build a very old verison of `bar` that doesn't work for some reason.
```
Building foo-X from CHaP and making a revision assumes you have already released `foo-X`.

README.md
```diff
@@ -426,4 +426,29 @@ There are some scripts for dealing with this:
...
+# repeat this whenever you change the repository, e.g. by adding a revision
+> foliage build -j 0 --write-metadata
+> nix build 
+  --override-input CHaP /home/user/cardano-haskell-packages/_repo
+  .#"ghc926/foo/X"
+```
+
+There are then two ways to make progress:
+1. Make a revision to rule out the bad case. If your build fails because `baz-Z` can't build with `bar-P`, then `baz-Z` _should_ have a constraint that excludes `bar-P`. You can add this constraint by making a revision to `baz-Z` and try again.
+2. Add a constraint to force the good case. If you are expecting to build with `bar-Y`, you can add a `bar == Y` constraint and the solver should tell you why it's impossible! You can either add this to the `cabal.project` specified in `nix/builder.nix`, or add `.addConstraint "bar == Y"` to the nix invocation.
```
andreabedini, line 448: I'd suggest to do this first, so you understand where the issue is. If you make a revision to force a new `bar` you might end up with no valid plan at all. We should warn the user about this possibility.

andreabedini (commented):
I left a couple of comments :+1:

README.md
michaelpj: I think this approach works whether or not you have a solver error for no plan or the wrong plan?

README.md
michaelpj: "released" is ambiguous, I'm thinking of when you're making a PR to add it but the CI says no

PR #332, opened 2023-06-07
Improve scripts/add-revions.sh

hamishmack:
The need to pass a `BUILT_REPO` arg is very confusing and it is small enough to just DL the latest to a temp dir.

This change also only uses the `BUILT_REPO` if there is no existing revision.  If there is a revision it copies the `.cabal` file from there (in case it is not in the BUILT_REPO yet).

Also uses `date -u` instead of `date --utc` so that we don't have to install `gdate` on macOS to make it work.


hamishmack:
If we need to pass in a different `_repo` we could make something like this work:

```
BUILT_REPO=_repo .scripts/add-revions.sh foo 1.0.0
```

scripts/add-revision.sh
```diff
@@ -28,15 +28,18 @@ done
...
 
-if ! shift 3; then
+if ! shift 2; then
   usage
   exit 1
 fi
 
+BUILT_REPO=$(mktemp -d)
+mkdir -p $BUILT_REPO/index
+curl -L https://input-output-hk.github.io/cardano-haskell-packages/01-index.tar.gz | tar -C $BUILT_REPO/index -xz
```
andreabedini, line 41: One can do `curl -L https://input-output-hk.github.io/cardano-haskell-packages/01-index.tar.gz | tar xvz cardano-cli/8.0.0/cardano-cli.cabal` to get the latest revision for `cardano-cli-8.0.0`.

andreabedini: Or even
```
curl -L https://input-output-hk.github.io/cardano-haskell-packages/package/cardano-cli-8.0.0/cardano-cli.cabal
```

andreabedini (dismissed):
LGTM

michaelpj:
This won't work now if you need to add a revision to something that's not in the remote, or if you're doing multiple revisions to something. That's uncommon, but kind of important when you need it. We could wrap this in something that fetches the repository, it would be nice to keep the bare one also.

scripts/add-revision.sh
michaelpj: Is this even correct? This downloads only the index. I guess that probably works in this case but I'm not sure. I think it would be safer to get the entire latest repository using one of the methods in https://github.com/input-output-hk/cardano-haskell-packages/pull/323

aniketd:
Changes to `use-gnu-tar.sh` as well as `current-timestamp.sh` from this one still seem relevant.

PR #335, opened 2023-06-08
use self-hosted runners for github actions

dermetfan:
This also removes nixbuild.net as I guess the point of self-hosted runners is to build ourselves, too. Also their action [writes system-wide SSH config](https://github.com/nixbuild/nixbuild-action/blob/2c22043033fd4118ff8f465d59e1edb7619fca74/nixbuild-action.sh#L123) which is hard to work around.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,44 +10,25 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  NIX_CONFIG: accept-flake-config = true
+
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: nixos
 
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # the check script below needs the whole history
 
-      - name: Setup Nix
-        uses: cachix/install-nix-action@v20
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable
-          extra_nix_config: accept-flake-config = true
-
-      - name: Setup nixbuild.net
-        uses: nixbuild/nixbuild-action@v16
-        with:
-          nixbuild_token: ${{secrets.NIXBUILD_TOKEN}}
-
       - name: Run checks
         run: nix develop -c ./scripts/check.sh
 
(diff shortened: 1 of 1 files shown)


disassembler (approved):

angerman:
I'm not sure we want to disable nixbuild?

michaelpj:
We absolutely do not want to disable nixbuild! Theses builds take a _long time_, and we need the runners.

michaelpj:
I'm not super pleased that this got merged without a review from someone who knows the CI well.

andreabedini:
I have to agree with @michaelpj, a lot of work went into this CI setup. I was suprised to see it reverted without a discussion.

dermetfan:
If nixbuild is necessary, self-hosted runners are not an option at all and this has to be reverted entirely.

michaelpj:
Because of the SSH config issue? We can ask them to see if they can improve that. What specifically is the problem?

dermetfan:
Yes. It tries to [write `/etc/ssh/ssh_config`](https://github.com/nixbuild/nixbuild-action/blob/2c22043033fd4118ff8f465d59e1edb7619fca74/nixbuild-action.sh#L120-L123) with sudo:

```
# Append ssh config to system config
sudo mkdir -p /etc/ssh
sudo touch /etc/ssh/ssh_config
sudo tee -a /etc/ssh/ssh_config < "$SSH_CONFIG_FILE" >/dev/null
```

They probably only tested on github-provided runners.

dermetfan:
Well there are possible workarounds, although quite a bit more work than just enabling github runners in the NixOS config:

- Do not use the nixbuild action and instead configure nixbuild on the host directly.
- Spawn containers or VMs so we can use sudo inside and run the github runner in them.

angerman:
How about this:
We designate ci8 as a dedicated shared runner for now.
Add the nixbuild ssh key by hand.

And then figure out a more sustainable route going forward?

We are seemingly exceeding the free GH runners for eval purposes, so we'll need Self hosted Runner. We also want to use nixbuild.

PR #339, opened 2023-06-09
CI: add back nix build flags for nixbuild.net

dermetfan:
Partial rollback of #335 following https://github.com/input-output-hk/cardano-haskell-packages/pull/335#issuecomment-1584123789.

nixbuild.net is now configured system-wide on the self-hosted runners.


angerman:
I think we can close this for now.

PR #355, opened 2023-06-20
Added optparse-applicative-0.16.1.1

erikd:
From https://github.com/input-output-hk/optparse-applicative at ac9a7ac5a872e2348c38cda928fce614a311f229


michaelpj:
- There is no issue for this
- It does not follow our versioning convention for patched packages

PR #356, opened 2023-06-20
Release 8.2 cardano packages

fraser-iohk:


fraser-iohk:
missed that there was already a set of releases for 8.2 by @teodanciu, so closing

PR #367, opened 2023-07-04
coot/cardano client 0.1.0.3

coot:
- Added cardano-client-0.1.0.3
- byron-spec-chain-0.1.1.2 revision


lehins (approved):
Thank you.
We'll take care of the actual compatibility with hedgehog-1.3 a bit later: https://github.com/input-output-hk/cardano-ledger/issues/3514

PR #385, opened 2023-07-18
Tar built repository before uploading it as artifact

### LINKED ISSUE #165: Try using upload/download tarifact to speed up the repository transfer in CI
https://github.com/actions/upload-artifact/issues/36#issuecomment-1129688501

Currently uploading and downloading both take ~1m, and it'll only get worse as the repository gets bigger. Sounds like having lots of files is a problem, which we do because of the website mostly.

andreabedini:
Closes #165


michaelpj:
Great, thanks.

PR #423, opened 2023-08-02
Added int-cast-0.2.0.0.0.0.0.0.2

jasagredo:
From https://github.com/input-output-hk/int-cast at a95a5eb8d3a4633a2230dd058b38b7ae82dd19a2

Compatible with GHC 9.6


michaelpj:
Why do you think we need this? We fixed this in `plutus` by removing the dependency.

jasagredo:
Ah I think it was due to a combination of situations, some other package was now being picked up due to other constraints, which pulled some old plutus in, which pulled this in. I can build without it.

Actually it is still pulled in (I could try to see who is responsible for it), but it uses the hackage revisioned one https://hackage.haskell.org/package/int-cast-0.2.0.0/revisions/

PR #431, opened 2023-08-07
Add script to reproduce GitHub Actions runs locally

andreabedini:
This scripts should help troubleshooting CI failures when adding new packages.

Given a run id, the script does the following:
1. It downloads the built-repo artifact from the CI
2. It unpacks it in a temporary folder
3. It obtains the commit hash that triggered the run
4. It runs the same nix build; i.e.

```
nix build \
        "github:$REPO/${headSha}#allPackages" \
        --override-input CHaP "path:$ARTIFACT_TEMP_DIR/_repo" \
        --accept-flake-config true \
        --print-build-logs --show-trace
```

It is also possible to specify which derivation to build.

The scripts uses the GitHub CLI utility `gh` because one needs to authenticate to download artifacts. I added `gh` to the dev shell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -155,6 +155,7 @@
               coreutils
               curlMinimal.bin
               gitMinimal
+              gh                  # GitHub CLI tool
               gnutar
               foliage.packages.${system}.default
             ];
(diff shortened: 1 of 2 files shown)


andreabedini:
> Doesn't look like it gets moved into the repo folder which I think you need, though?

At the moment yeah, but it's there if we need to use it.

I don't know. I made this script as @disassembler requested but I am not sure it fits the use-case.

PR #446, opened 2023-08-21
Consensus: mark private sublibs as unbuildable

amesgen:
Alternative to #445, see https://github.com/input-output-hk/cardano-haskell-packages/pull/445#issuecomment-1685878155

Revision diff is just adding the respective `else buildable: False` branches; will be mirrored in Consensus if this PR is merged.


michaelpj:
Note you only need to do this in libraries that _depend_ on the private sublibraries, so you could skip e.g. `ouroboros-consensus-cardano`.

amesgen:
https://github.com/input-output-hk/cardano-haskell-packages/pull/445#issuecomment-1686326396

PR #455, opened 2023-08-25
Patch conway and shelley packages to fix DRep stake distribution

lehins:
Minor releases of two packages: https://github.com/input-output-hk/cardano-ledger/pull/3677

### DIFF
--- a/_sources/cardano-ledger-conway/1.7.1.0/meta.toml
+++ b/_sources/cardano-ledger-conway/1.7.1.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2023-08-25T10:29:18Z
+github = { repo = "input-output-hk/cardano-ledger", rev = "81d2e208a4509f89b8602973c47aaf572da8dd77" }
+subdir = 'eras/conway/impl'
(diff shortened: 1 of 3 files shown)


teodanciu:
Looks good to me, except I can't understand why this compilation error in `build-new-packages` is happening:

  > src/Cardano/Ledger/Conway/Governance/Procedures.hs:374:7: error:
       >     parse error on input ‘!’
       >     |
       > 374 |       !(StrictMaybe (PrevGovActionId 'PParamUpdatePurpose (EraCrypto era)))
       >     |       ^

lehins:
> Looks good to me, except I can't understand why this compilation error in build-new-packages is happening:

Looks like haddock was enabled recently in CI: https://github.com/input-output-hk/cardano-haskell-packages/pull/441

If my assumption is correct, until recently haddock had a bug where you could not add documentation to fields of a non-record type, eg. only newer haddock can handle this:

```
data Foo = 
  Foo
  -- | This is Int
  Int
  -- This is Bool
  Bool
```

michaelpj:
My understanding is that this will break anyone trying to get an 8.10 shell including the ledger documentation. If only we could drop 8.10...

PR #467, opened 2023-09-04
Clean up version flags in add-from-github

michaelpj:
- The flag to force the version was appealingly named `-v` and looked like you were just supposed to always pass the version. I made it scary and changed the flag so that anyone who is currently using it gets tripped up
- The "revision" flag is extremely confusing given we also have, you know, revisions. It's also not used that much, so I just removed it.

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -6,31 +6,26 @@ set -o pipefail
 SCRIPT_DIR=$(dirname "$(which "$0")")
 
 function usage {
-  echo "Usage $(basename "$0") [-r REVISION] [-v VERSION] REPO_URL COMMIT-SHA [SUBDIRS...]"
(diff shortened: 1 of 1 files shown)


angerman:
This seems sensible.

angerman:
This makes me wonder if we want to have a GitHub action that shellchecks our shell scripts?

michaelpj:
By all means...

michaelpj:
Also personally I would prefer it if we made foliage sub-commands for these operations...

angerman:
> Also personally I would prefer it if we made foliage sub-commands for these operations...

@michaelpj you mean all the shell scripts to be turned into `foliage cmd ...`?

michaelpj:
Yes exactly. Foliage knows how to do all this stuff correctly already.

PR #470, opened 2023-09-05
Add shellcheck to CI

angerman:
This should help ensure we have our shell scripts checked and changes won't introduce unexpected behaviour.

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -64,12 +64,12 @@ if [[ ${#SUBDIRS[@]} -gt 1 && (-n $VERSION || -n $REVISION) ]]; then
   exit 1
 fi
 
-if [[ ! "$REPO_URL" =~ "https://github.com/" ]]; then
+if [[ ! "$REPO_URL" =~ https://github.com/ ]]; then
   echo "Provided url is not a github url: $REPO_URL"
   exit 1
 fi
 
-TIMESTAMP=$($SCRIPT_DIR/current-timestamp.sh)
+TIMESTAMP=$("$SCRIPT_DIR"/current-timestamp.sh)
 
 render_meta() {
   local TIMESTAMP=$1
@@ -145,7 +145,7 @@ do_package() {
   git commit -m"Added $PKG_ID" -m "From $REPO_URL at $REPO_REV"
 }
 
-WORKDIR=$($SCRIPT_DIR/fetch-github-cabal-files.sh $REPO_URL $REPO_REV)
+WORKDIR=$("$SCRIPT_DIR"/fetch-github-cabal-files.sh "$REPO_URL" "$REPO_REV")
 
 if [[ ${#SUBDIRS[@]} -eq 0 ]]; then
   do_package "$REPO_URL" "$REPO_REV" "" "$WORKDIR"
--- a/scripts/check-archive-extension.sh
+++ b/scripts/check-archive-extension.sh
@@ -3,6 +3,7 @@
 # Use gnu-tar and gnu-date regardless of whether the OS is Linux
 # or BSD-based.  The correct command will be assigned to TAR and DATE
 # variables.
+# shellcheck disable=SC1091
 source "$(dirname "$(which "$0")")/use-gnu-tar.sh"
 
 # for some reason you can't just put a heredoc in a variable...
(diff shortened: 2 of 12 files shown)


scripts/add-from-github.sh
```diff
@@ -64,12 +64,12 @@ if [[ ${#SUBDIRS[@]} -gt 1 && (-n $VERSION || -n $REVISION) ]]; then
   exit 1
 fi
 
-if [[ ! "$REPO_URL" =~ "https://github.com/" ]]; then
+if [[ ! "$REPO_URL" =~ https://github.com/ ]]; then
```
angerman, line 67: Are we sure about this @erikd? @andreabedini do you know?

This was one of the few I was unsure about when reading spellcheck logs, due to the involvement of regex and `/`.

erikd: That was the one I was also unsure of, but `shellcheck` recommended it

andreabedini:
```
~
❯ { REPO_URL=https://github.com/input-output-hk/cardano-haskell-packages; [[ ! "$REPO_URL" =~ https://github.com/ ]]; echo $? }
1

~
❯ { REPO_URL=https://github.com/input-output-hk/cardano-haskell-packages; [[ ! "$REPO_URL" =~ "https://github.com/" ]]; echo $? }
1
```
it seems to workm the same?

scripts/version-lookup.sh
```diff
@@ -3,18 +3,18 @@
...
 
 versions() {
-  cat dist-newstyle/cache/plan.json \
-    | jq -r '
+  jq -r '
       ."install-plan"[]
       | select(."pkg-src".repo.uri == "https://input-output-hk.github.io/cardano-haskell-packages")
       | {"pkg-name": ."pkg-name", "pkg-version": ."pkg-version"}
-      | @base64' \
+      | @base64' < dist-newstyle/cache/plan.json \
```
angerman, line 17:
```suggestion
      | @base64' \
      dist-newstyle/cache/plan.json \
```
wouldn't that work just fine? One less redirection to read?

scripts/version-lookup.sh
```diff
@@ -23,5 +23,5 @@ for x in $(versions); do
   pkg_json="$(echo "\"$x\""  | jq -r '@base64d')"
   pkg_name="$(echo "$pkg_json" | jq -r '."pkg-name"')"
   pkg_version="$(echo "$pkg_json" | jq -r '."pkg-version"')"
-  cat "$BASE_DIR/_sources/$pkg_name/$pkg_version/meta.toml" | yj -tj | jq "{\"root\": (. + {\"package\": {\"name\": \"$pkg_name\", \"version\": \"$pkg_version\"}})}"
+  yj -tj < "$BASE_DIR/_sources/$pkg_name/$pkg_version/meta.toml" | jq "{\"root\": (. + {\"package\": {\"name\": \"$pkg_name\", \"version\": \"$pkg_version\"}})}"
```
angerman, line 26: I think this also works with `yj -tj FILE`?

scripts/list-latest-package-versions.sh
```diff
@@ -4,4 +4,4 @@
 
 REPO=$1
 
-cat "$REPO/foliage/packages.json" | jq -r 'group_by(."pkg-name") | .[] | max_by(."pkg-version") | [."pkg-name", ."pkg-version"] | @tsv' 
+jq -r 'group_by(."pkg-name") | .[] | max_by(."pkg-version") | [."pkg-name", ."pkg-version"] | @tsv' < "$REPO/foliage/packages.json"
```
angerman, line 7:
```suggestion
jq -r 'group_by(."pkg-name") | .[] | max_by(."pkg-version") | [."pkg-name", ."pkg-version"] | @tsv' "$REPO/foliage/packages.json"
```

scripts/check-archive-extension.sh
```diff
@@ -57,6 +59,7 @@ NEW_LISTING=$("$TAR" -tvf "$NEW_INDEX_TAR" | sort -k4,5)
 
 DIFF=$(diff <(echo "$OLD_LISTING") <(echo "$NEW_LISTING"))
 
+# shellcheck disable=SC2181
```
angerman, line 62: I think this warrants a comment when disabling shellcheck. Why do we disable it here?

scripts/add-from-github.sh
angerman: Good!

scripts/version-lookup.sh
erikd: No, that will not work. `jq` is weird and by default works by reading input from `stdin`.

scripts/list-latest-package-versions.sh
erikd: Nope that is not correct. See above.

scripts/version-lookup.sh
erikd: I do not have `yj` installed locally so I cannot test this, but I do know `yj` is very similar to `jq` in many ways, and `jq` would not work correctly like that.

scripts/check-archive-extension.sh
erikd: Added a full comment there.

angerman:
```
jq FILTER FILE
```
is equivalent to `cat FILE | jq FILTER`, 
however
```
yj FLAGS FILE
```
is _not_ accepted.

Let's stick with `< FILE` for consistency.

PR #485, opened 2023-09-07
GHC 9.6 CI

michaelpj:
Let's observe the carnage and see if we can start enabling this.


andreabedini (approved):

michaelpj:
Looks like it's really just the node in the smoke test packages, which is pretty great.

PR #490, opened 2023-09-07
coot/consensus release

coot:
- Added sop-extras-0.1.0.0
- Added ouroboros-consensus-cardano-0.9.0.0
- Added strict-sop-core-0.1.0.0
- Added ouroboros-consensus-0.11.0.0
- ouroboros-consensus-protocol-0.5.0.6 revision


coot:
The `build-packages` CI fails, because `cardano-api` cannot build against the changes in the released packages.  Either we didn't deal well with package version on consensus side or `cardano-api` has to relaxed bounds (e.g. missing upper bounds).

nfrisby:
Yes, I think `cardano-api` needs a `ouroboros-consensus < 0.11` constraint, because the SOP modules were moved out into a separate `sop-strict-core` package.

michaelpj:
Don't forget to push the constraints to the package definition!

PR #499, opened 2023-09-14
Add marlowe-plutus-0.1.0.0

jhbertra:
- [x] Adds marlowe-plutus package
- [x] Adds marlowe-runtime as a smoke test package.
      - This was breaking frequently because of improperly constrained dependency versions in upstream packages, as such it is an excellent smoke test to detect breakages.


michaelpj:
> Adds marlowe-runtime as a smoke test package

:+1:

jhbertra:
Removed marlowe-runtime smoke test because the build was exceeding 4 hours.

michaelpj:
That's not necessarily a problem: it should all get cached :thinking:

michaelpj:
Added in https://github.com/input-output-hk/cardano-haskell-packages/pull/500

PR #516, opened 2023-09-27
Update Marlowe packages

jhbertra:


flake.nix
```diff
@@ -76,36 +76,40 @@
           ghc92.enabled = false;
           ghc96.enabled = false;
         };
+        hasql-dynamic-syntax = {
+          ghc810.enabled = false;
+          ghc96.enabled = false;
+        };
         marlowe-cardano = {
-          ghc92.enabled = false;
+          ghc810.enabled = false;
```
michaelpj, line 84: You know that the official supported version for the node is still 8.10, right? Are you sure you want to stop building with 8.10?

jhbertra: Yes, I'm sure. Building with GHC 8.10 would prevent us from taking advantage of newer language features and library versions that require base 4.16. We aren't a dependency for node, so us being on a higher GHC version isn't a concern if the wire protocol is compatible and `cardano-api` builds on 9.2.

PR #536, opened 2023-10-16
Kwxm/plutus 1.15

kwxm:
Plutus release 1.15


zliu41 (changes requested):
We use four component version numbers, i.e. 1.15.0.0.

kwxm:
> We use four component version numbers, i.e. 1.15.0.0.

Damn.  I think I''ll have to start again from the beginning.

PR #551, opened 2023-10-30
Rework the CI

andreabedini:
This reworks the CI to give more control to GitHub Actions.

- The flake exposes the list of smoke-test derivations that GHA load and execute.
- The repositories for base (available only for PRs) and head are built concurrently
- Built repositories are cached by their commit hash, we should not need to build them more than once. We can always delete the specific cache in case of issues.

Minor changes:
- haskellPackages -> chapPackages
- Consistent camelCase
- chapPackagesLatestVersion maps to a single version rather than a list with a single element

Opening as draft to see if it works.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
   check:
     runs-on: nixos
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0 # the check script below needs the whole history
 
@@ -90,6 +90,8 @@ jobs:
       - build-repo-base
       - build-repo-head
     steps:
+      - uses: actions/checkout@v4
+
       - name: Download built repository (base)
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -14,218 +14,235 @@ env:
...
+      - build-repo-base
+      - build-repo-head
+      - generate-smoke-tests
+    strategy:
+      fail-fast: false
+      # The step generate-smoke-tests produces a list of derivations to build
+      # include:
+      #  - name: ghc96/cardano-node/8.5.0
+      #  ...
+      matrix: ${{ fromJSON(needs.generate-smoke-tests.outputs.matrix) }}
```
michaelpj, line 148: This is not a good idea, I'm afraid. The reason is that there is a cap on the maximum number of jobs that a matrix can generate :( I ran into this trying to do something like this for HLS. I think we would hit the cap quite fast (it's ~500). We could potentially do one job per GHC version I guess.

michaelpj (commented):
Just a comment on the main approach. Also, we did something like this in the past, I think, and I moved it to just build the smoke test packages derivation because we wanted to do less in GHA. But I do think it can be nice to do _more_ in GHA if we get a better UI from it.

PR #568, opened 2023-11-16
Revise fs-api and fs-sim for newer io-classes

jasagredo:


jorisdral (approved):
Are you updating the bounds in the `fs-sim` repo too?

PR #569, opened 2023-11-16
Add 9.8 to CI, remove 9.2 and 9.4

michaelpj:
We're moving towards 9.8, we're not going to use 9.2 or 9.4 in the long run.

Since 9.8 is only going to work for a small set of packages to begin with, I added an `experimental-compilers` list. Things in that list are off by default but you can turn them on in `exceptions`.

`cardano-prelude` builds with 9.8 now, so that's our first test!

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -139,11 +139,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1692799911,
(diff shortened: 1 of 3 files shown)


michaelpj:
I put 9.2 back, since some teams are still using  it, and this would otherwise leave some package with no supported GHC versions!

PR #580, opened 2023-11-21
Bump base upper bounds of some cardano-base packages (needed for ghc-9.8).

erikd:


michaelpj:
Is this enough? I see that the corresponding [PR](https://github.com/input-output-hk/cardano-base/pull/438/files) had some code changes also?

erikd:
> Is this enough? I see that the corresponding [PR](https://github.com/input-output-hk/cardano-base/pull/438/files) had some code changes also?

In `cardano-crypto-class` which is not included here because a new release has already been done.

PR #583, opened 2023-11-22
Publish 1.5.0.2 for Plutus Core

zliu41:


michaelpj:
I'm going to force-merge this: `plutus-1.5.0.2` won't build with the later GHCs that we try to build it with, and we don't have a good way to do per-package-version exceptions right now.

PR #584, opened 2023-11-23
Update CHaP pin

michaelpj:


andreabedini:
I merged https://github.com/input-output-hk/cardano-haskell-packages/pull/589 which seemed to be a more recent version of the same.

PR #633, opened 2024-01-02
Added hedgehog-extras-0.5.0.0

smelc:
From https://github.com/input-output-hk/hedgehog-extras at 1ec284b5e728e2beabda7512a5e76eae8b667ebb


carbolymer:
we're using hedgehog-extras from hackage, not from chap

carbolymer (changes requested):
use hackage version, not chap

PR #635, opened 2024-01-05
cardano-node-emulator-1.3.0.0

sjoerdvisscher:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -181,7 +181,9 @@
             {
               # Packages that depend on the plutus-tx plugin have broken haddock
               packages = {
+                cardano-node-emulator.doHaddock = false;
                 plutus-ledger.doHaddock = false;
+                plutus-script-utils.doHaddock = false;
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -145,11 +147,15 @@
           ghc92.enabled = false;
           ghc96.enabled = false;
         };
-        quickcheck-contractmodel = {
-          ghc96.enabled = false;
+        plutus-ledger = {
+          ghc810.enabledVersions = [ "1.2.0.0" ];
+          ghc92.enabledVersions = [ "1.3.0.0" ];
+          ghc96.enabledVersions = [ "1.3.0.0" ];
```
michaelpj, line 153: This could get pretty tedious. Perhaps we could instead have `ghc810.enabledPredicate = v : ...`? Then you can use stuff like `builtins.compareVersions` to state the constraints more compactly. And we can replace `enabled` by `enabledPredicate = v : true`

michaelpj (approved):
Looks great!

PR #648, opened 2024-01-28
Bump base upper bound on non-integral

erikd:

### DIFF
--- a/_sources/non-integral/1.0.0.0/revisions/2.cabal
+++ b/_sources/non-integral/1.0.0.0/revisions/2.cabal
@@ -1,0 +1,42 @@
+cabal-version:      3.0
+name:               non-integral
+version:            1.0.0.0
+license:            Apache-2.0
+maintainer:         operations@iohk.io
+author:             IOHK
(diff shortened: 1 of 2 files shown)


_sources/non-integral/1.0.0.0/revisions/2.cabal
```diff
@@ -0,0 +1,42 @@
...
+
+library
+    exposed-modules:  Cardano.Ledger.NonIntegral
+    hs-source-dirs:   src
+    default-language: Haskell2010
+    ghc-options:
+        -Wall -Wcompat -Wincomplete-record-updates
+        -Wincomplete-uni-patterns -Wredundant-constraints -Wunused-packages
+
+    build-depends:    base >=4.14 && <4.20
```
lehins, line 26: You've set it on master to:
```suggestion
    build-depends:    base >=4.14 && <5
```

Just to avoid confusion in the future it would be nice to have this revision reflect the same state as on master

lehins (approved):
Looks good. One small suggestion.

PR #650, opened 2024-01-29
Update core marlowe libraries

jhbertra:


michaelpj:
It's meant to just be these two spec ones?

jhbertra:
Yes, the cardano library upgrades will come later and depend on these.

PR #663, opened 2024-02-14
Update `cardano-base` packages

Lucsanszky:

### DIFF
--- a/_sources/cardano-ledger-shelley/1.0.0.0/meta.toml
+++ b/_sources/cardano-ledger-shelley/1.0.0.0/meta.toml
@@ -1,3 +1,7 @@
 timestamp = 2023-02-23T15:33:46Z
 github = { repo = "input-output-hk/cardano-ledger", rev = "2f5956038233e4df0a065d96db32398605603f9b" }
 subdir = 'eras/shelley/impl'
+
+[[revisions]]
+number = 1
+timestamp = 2024-02-15T15:59:55Z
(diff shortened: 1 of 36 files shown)


Lucsanszky:
@lehins, should we also release some newer versions of other packages? It seems like `cardano-mempool` was never released and `cardano-crypto-praos` has some funkiness going on with its versions (on CHaP, the current version is `2.1.2.0` but the current version according to the `.cabal` file is `2.1.1.2`). `cardano-crypto-class` also has a newer version but the `CHANGELOG` states that we shouldn't release that yet so I won't.

lehins:
> should we also release some newer versions of other packages?

@Lucsanszky No. As you did notice in the changlog, those packages are not ready be released yet.

https://github.com/IntersectMBO/cardano-base/blob/f11ddc7f87038c16d5a6d7611fb5dda51be9533d/cardano-crypto-class/CHANGELOG.md?plain=1#L5-L6

lehins (changes requested):
This looks good, but we need to fix all of the versions of cardano-ledger-shelley from 1.0.0.0 to 1.9.0.0, not just 1.2.0.0 and 1.9.0.0 since we know that they are too incompatible with `cardano-slotting-0.2`. 
The fact that CI didn't pick up on them doesn't mean those versions would not break if they got picked up by cabal solver. Since we know about this incompatibility, we should do the due diligence and prevent such build plan to ever exist.

So, yeah, just and adding a whole bunch more revisions like that and we'll be good with the PR :slightly_smiling_face:

lehins (approved):
Perfect! Thank you.

PR #672, opened 2024-02-22
strict-checked-vars revision

bolt12:

### DIFF
--- a/_sources/strict-checked-vars/0.2.0.0/revisions/1.cabal
+++ b/_sources/strict-checked-vars/0.2.0.0/revisions/1.cabal
@@ -1,0 +1,99 @@
+cabal-version:   3.0
+name:            strict-checked-vars
+version:         0.2.0.0
+synopsis:
+  Strict MVars and TVars with invariant checking for IO and IOSim
+
+description:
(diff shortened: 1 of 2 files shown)


_sources/strict-checked-vars/0.2.0.0/revisions/1.cabal
```diff
@@ -0,0 +1,99 @@
...
+  exposed-modules:
+    Control.Concurrent.Class.MonadMVar.Strict.Checked
+    Control.Concurrent.Class.MonadSTM.Strict.TVar.Checked
+
+  default-language: Haskell2010
+  build-depends:
+    , base         >=4.9 && <5
+    , io-classes   >=1.2 && <1.5
+    , strict-mvar  >=1.2 && <1.4
+    , strict-stm   >=1.2 && <1.4
```
amesgen, line 55: These also need to be bumped, right?

bolt12: true!

amesgen (approved):
I will also open a corresponding PR mirroring this change in `cardano-base` :+1:

PR #681, opened 2024-02-26
Bump Plutus deps to 1.22.1.0

Unisay:


michaelpj (approved):

michaelpj:
We're going to deprecate 1.22.0. Perhaps we should just go to 1.23, I'm not sure :thinking:

lehins:
> We're going to deprecate 1.22.0.

Interesting. Is there support in CHaPs for deprecating packages?

michaelpj:
Yes. I need to add a section to the readme but see an example here: https://github.com/IntersectMBO/cardano-haskell-packages/blob/44de6cd146e40fa8d5ee7f3f0e0e2af76e2f0833/_sources/diff-containers/1.1.10/meta.toml#L6

michaelpj:
Note that this does "cabal-style deprecations", which is to say it makes them non-preferred, so the solver _can_ still pick them if it has no other options (I wish it emitted a warning in this case...).

Unisay:
I've added the deprecations for 1.22.0.0

PR #690, opened 2024-03-06
Fix consensus node 8.9

jasagredo:
Backports of IntersectMBO/ouroboros-consensus#983 and IntersectMBO/ouroboros-consensus#985


jasagredo:
Should not be merged until https://github.com/IntersectMBO/ouroboros-consensus/pull/986 is green (except maybe the Changelog check)

PR #698, opened 2024-03-15
Added cardano-client-0.3.1.2

bolt12:
From https://github.com/intersectmbo/ouroboros-network at 624085b7a84f099b3f669cda61c380e87fd5d0a7


bolt12:

PR #736, opened 2024-04-08
Release ouroboros-consensus-diffusion-0.14.0.0

amesgen:
Note that this is a **backport release**, see https://github.com/IntersectMBO/ouroboros-consensus/pull/1042, 0.13.0.0 is still the latest and greatest. (We might change our versioning scheme in the future to make that more clear.)

### DIFF
--- a/_sources/ouroboros-consensus-cardano/0.14.2.0/revisions/2.cabal
+++ b/_sources/ouroboros-consensus-cardano/0.14.2.0/revisions/2.cabal
@@ -1,0 +1,631 @@
+cabal-version:   3.0
+name:            ouroboros-consensus-cardano
+version:         0.14.2.0
+synopsis:
+  The instantation of the Ouroboros consensus layer used by Cardano
+
+description:
+  The instantation of the Ouroboros consensus layer used by Cardano.
+
+license:         Apache-2.0
+license-files:
+  LICENSE
+  NOTICE
+
+copyright:
+  2019-2023 Input Output Global Inc (IOG), INTERSECT 2023-2024.
+
+author:          IOG Engineering Team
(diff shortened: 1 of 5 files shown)


_sources/ouroboros-consensus-cardano/0.14.2.0/revisions/2.cabal
```diff
@@ -0,0 +1,631 @@
...
+    , filepath
+    , fs-api                         ^>=0.2
+    , githash
+    , microlens
+    , mtl
+    , network
+    , nothunks
+    , ouroboros-consensus            ^>=0.16
+    , ouroboros-consensus-cardano
+    , ouroboros-consensus-diffusion  ^>=0.14
```
coot, line 529: Do you want to allow `^>=0.12` as well?  Won't this change break `cardano-node`?

coot (commented):
LGTM

_sources/ouroboros-consensus-cardano/0.14.2.0/revisions/2.cabal
amesgen: Hmm, it shouldn't break the node as it uses `index-state` which controls revisions, but it also shouldn't hurt. Will change it if I need to make another change to make CHaP CI happy.

amesgen: Ah you mean here in CHaP :facepalm: CI indeed fails, pushed a fix.

coot (approved):
LGTM

amesgen:
I am going to merge this, #1042 is only waiting for x86_64-darwin CI (aarch64-darwin already passed).

PR #737, opened 2024-04-08
ouroboros-consensus-cardano: relax cardano-git-rev bounds

coot:
* cardano-api: relaxed bounds of ouroboros-consensus-diffusion
* ouroboros-consensus-cardano: relax cardano-git-rev bounds


amesgen (changes requested):
I think the oc-cardano bounds for cardano-git-rev should stay as is, oc-cardano-0.14.1.0 only compiles with cardano-git-rev-0.1.*, and oc-cardano-0.14.2.0 only with cardano-git-rev-0.2.*.

I think/hope that node 8.9.2 can use oc-cardano-0.14.1.0.

amesgen (approved):
:+1: on the oc-cardano-0.14.1.0 revision (allowing oc-diffusion-0.14)

PR #773, opened 2024-05-23
Force Cabal to be in the plan

andreabedini:
This tries to reduce the divergences between Haskell.nix and
cabal-install, as they can cause issues (e.g. Haskell.nix reinstall ghc
by default, assuming that Cabal is in the plan, which it might not be).


andreabedini:
Closed in favour of https://github.com/IntersectMBO/cardano-haskell-packages/pull/772

PR #777, opened 2024-06-03
Added cardano-crypto-class-2.1.5.0

paluh:
From https://github.com/IntersectMBO/cardano-base at 994d5b4

### DIFF
--- a/_sources/cardano-crypto-class/2.1.5.0/meta.toml
+++ b/_sources/cardano-crypto-class/2.1.5.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2024-06-03T14:31:53Z
+github = { repo = "IntersectMBO/cardano-base", rev = "994d5b4" }
+subdir = 'cardano-crypto-class'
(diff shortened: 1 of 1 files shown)


_sources/cardano-crypto-class/2.1.5.0/meta.toml
```diff
@@ -0,0 +1,3 @@
+timestamp = 2024-06-03T14:31:53Z
+github = { repo = "IntersectMBO/cardano-base", rev = "994d5b4" }
```
lehins, line 2: Let's use the full sha for the release: 
```suggestion
github = { repo = "IntersectMBO/cardano-base", rev = "994d5b42e28b98e4a4a53607182b93b6390e5487" }
```

lehins (commented):
Other than using a sha prefix it looks good

paluh:
Oh, sorry - I used the prefix when invoked the helper script and somewhat missed that in the commit diff.

PR #788, opened 2024-06-13
Added unix-time-0.4.14.0.0.0.0.1

hamishmack:
From https://github.com/hamishmack/unix-time at 7c28e2294bc3b1a9cb7ec12d7d6696d74ffab48f

One character change to fix cross compilation of the unix-time package:

```
 #if defined(_WIN32)
-#include <Windows.h>
+#include <windows.h>
 #endif
```

Upstream PR is https://github.com/kazu-yamamoto/unix-time/pull/64


angerman (approved):
lgtm.

angerman:
As per https://github.com/kazu-yamamoto/unix-time/pull/64#issuecomment-2167017181, should we just wait up to 24hs, and not pollute CHaP with this?

PR #794, opened 2024-06-19
Bump base dependency in cardano-crypto-praos-2.1.2.0

erikd:
From https://github.com/IntersectMBO/cardano-base at a9bfdf50b7794c962f73f06763546dc65257720e


lehins (approved):
Looks good. Thank you!

PR #801, opened 2024-06-24
Added cuddle-0.1.22.0

nc6:
From https://github.com/input-output-hk/cuddle at be1e3d04f2cffb006e2fbb5cd8aeba266de29060

This is a library for defining and validating CDDL


teodanciu (commented):
Just echoing Alexey's question about the base bounds `   base  ^>=4.14.3.0 || ^>=4.16.3.0 || ^>=4.18.1.0 || ^>=4.19.0.0` - I know this is technically a comment for cuddle itself  than this PR, but wondering  if  it won't  make it more difficult to  use it if we release it with these bounds?

nc6:
It shouldn't make things more difficult. Those bounds just reflect the recent versions of GHC that have been released.

teodanciu (approved):
Looks good to me

PR #833, opened 2024-07-15
Update consensus code owners

jorisdral:


amesgen (approved):

jorisdral:
I'm not sure who can change the settings on this repository, but could it be possible to automatically request reviews from code owners? I don't think we should *require* an approval from a code owner

PR #874, opened 2024-08-22
Ledger release for cardano-node-9.2 (test suites)

Jimbo4350:
Related: https://github.com/IntersectMBO/cardano-haskell-packages/pull/872


lehins (approved):
:+1: 

We needed couple of more packages to be released, that where missing from the original PR

PR #882, opened 2024-08-29
Consensus release for node 9.2

Jimbo4350:


jasagredo:
I would use `IntersectMBO/ouroboros-consensus@3da919ffc0ef0105bc51f6bbd8dd12e8dd6990c6` instead, because that is the merge commit on main.

jasagredo (approved):
Looks good for the consensus packages, but there is the issue for the nothunks version compatibility.

PR #914, opened 2024-10-04
Plutus core release 1.35.0.0

ramsay-t:


zliu41 (changes requested):
I think `rev` should point to the commit hash rather than the tag, since tags are mutable.

zliu41 (approved):

PR #920, opened 2024-10-09
Update haskell.nix to latest version

hamishmack:


teodanciu:
@hamishmack  Thank you for this, I"m merging it because it's solving my problem, and I would like to release the package that is failing without this change.

PR #925, opened 2024-10-11
Release cardano-ledger-byron

teodanciu:
From https://github.com/input-output-hk/cardano-ledger at 3b8090905773bc8990755576d9fa7a5edf07e80d

### DIFF
--- a/_sources/cardano-ledger-byron-test/1.5.2.0/meta.toml
+++ b/_sources/cardano-ledger-byron-test/1.5.2.0/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2024-10-11T20:28:29Z
+github = { repo = "IntersectMBO/cardano-ledger", rev = "0ba8e73c41847142e0bed09e09a8aa166fc10384" }
+subdir = 'eras/byron/ledger/impl/test'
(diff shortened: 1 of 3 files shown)


Lucsanszky:
Actually, we can add a revision instead of making a PR in `ledger` with the `Cabal >=3.12` constraint. That way, we can use the same commit hash for the release as in #916 (`0ba8e73c41847142e0bed09e09a8aa166fc10384`).

Lucsanszky:
@neilmayhew indicated that the `Cabal >=3.12` constraint 'hack' might be problematic due to the fact that GHC 8.10.7 is bundled with Cabal 3.2.1.0. Are we okay to add this revision anyway, until `haskell.nix` is updated in a non-breaking way? 

cc @disassembler

neilmayhew:
> the Cabal >=3.12 constraint 'hack' might be problematic due to the fact that GHC 8.10.7 is bundled with Cabal 3.2.1.0

I was wrong about that (which is just as well, otherwise the revision wouldn't have worked either). The constraint is OK because `nix/builder.nix` adds `--allow-boot-library-installs` to the Cabal project's `configureArgs`.

aniketd:
Pinging @teodanciu 🙂 noticed this one still neither merged nor closed. Should we close it?

PR #928, opened 2024-10-15
Added cardano-crypto-1.1.3

erikd:
From https://github.com/IntersectMBO/cardano-crypto at 790a3027a210ed3684ea9ffa1da504873d96b7de


erikd:
Abandoning this. Will need to switch `ledger`, `plutus` and almost certainly others from `crytonite` to `crypton` simultaneously.

PR #952, opened 2024-12-11
Release cardano crypto version 2.2

lehins:
A long overdue release of `cardano-crypto-class[tests]-2.2` and `cardano-crypto-praos-2.2`

This PR also adds revisions with bounds fixups. For plutus-1.37 and 1.38 upper bound is relaxed to allow this new version, while for all released versions of `cardano-ledger-binary` upper bound was added to prevent this version of crypto being used until a compatible version of this package is released.

This effectively allows for future version of ledger and other downstream packages to use new version of crypto packages, while preventing its usage with any existing releases.

Other revisions added to the PR:
* upper bound on `data-default-class <0.2`, since it has recently been decided that it is better been switched to use `data-default` and dependency was inverted with `data-default-0.8`
* upper bound `data-default <0.8`, for the above reason and because `cardano-ledger-core` historically provided an instance for `Bool` that now conflicts with the one provided in `data-default-0.8`
* upper bound `quickcheck-instances <0.3.32`, since it provides an `Arbitrary` instance for primitive `Vector`, which also conflicted with the orphan instance historically provided by `cardano-ledger-binary`

It also required a bump to `hackage.nix` and `haskell.nix`, which was brought over from #953

### DIFF
--- a/_sources/cardano-ledger-binary/1.3.4.0/revisions/1.cabal
+++ b/_sources/cardano-ledger-binary/1.3.4.0/revisions/1.cabal
@@ -123,7 +123,7 @@ library testlib
         prettyprinter,
         primitive,
         QuickCheck,
-        quickcheck-instances,
+        quickcheck-instances <0.3.32,
         tasty-hunit,
         random >=1.2,
         text,
--- a/_sources/cardano-ledger-binary/1.3.4.0/revisions/1.cabal
+++ b/_sources/cardano-ledger-binary/1.3.4.0/revisions/1.cabal
@@ -1,0 +1,198 @@
+cabal-version: 3.0
+name:          cardano-ledger-binary
+version:       1.3.4.0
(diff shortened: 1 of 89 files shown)


_sources/cardano-ledger-binary/1.3.4.0/revisions/1.cabal
```diff
@@ -0,0 +1,198 @@
...
+        iproute,
+        half,
+        hedgehog,
+        hspec,
+        hspec-core,
+        pretty-show,
+        prettyprinter,
+        primitive,
+        QuickCheck,
+        quickcheck-instances <0.3.32,
```
lehins, line 126: @hamishmack Here is the bound that should prevent this failure from happening: https://github.com/IntersectMBO/cardano-haskell-packages/pull/952/checks?check_run_id=34290940620

PR #953, opened 2024-12-11
Update haskell.nix

hamishmack:
This was merged before in #920, but it looks like it is not in `main`.


lehins:
Let's close this PR in favor of #952 because there are a whole bunch of revisions necessary to make the CI pass. 
I've cherry-picked your commit for haskell.nix update from this PR.

PR #973, opened 2025-01-07
Update flake inputs

neilmayhew:


neilmayhew:
This isn't going to fix the current Hydra problems, and is in any case done by a bot build action

PR #977, opened 2025-01-14
cardano-api-10.6.0.0 & cardano-api-gen-10.1.0.0

carbolymer:


Jimbo4350:
I disabled auto-merge because @neilmayhew needs to merge one more change into `cardano-api`'s master.

PR #979, opened 2025-01-16
Switch from cryptonite to crypton

erikd:
From https://github.com/input-output-hk/cardano-crypto at 790a3027a210ed3684ea9ffa1da504873d96b7de


erikd:
Abandoning this.  Incorrect solution.

PR #949, opened 2024-12-10
Add some packages from cardano-wallet to CHaP

abailly:
This will fix https://github.com/cardano-foundation/cardano-wallet/issues/4411

PR #996, opened 2025-01-29
Possible fix for #949

hamishmack:


hamishmack:
I'm not sure if disabling ghc 8.10 was actually necessary.  Rebasing the changes onto the latest CHaP did seem to fix a lot of the failures.

hamishmack:
This looks like incompatible versions https://github.com/IntersectMBO/cardano-haskell-packages/actions/runs/13043668079/job/36390587302?pr=996#step:7:1164

Does something need updating?

abailly:
Thanks for looking at that @hamishmack Do I understand correctly I would need to rebase the PR on a recent CHaP? 
Definitely can do that...

abailly:
Sorry, I don't understand where the problem comes from: is there a bound missing?

abailly:
OK, I think I understand: The bounds are set in the `cabal.project` hence not available in the `.cabal` files, so when trying to build with "whatever" versions are available it has a high probability of failing 🤦

abailly:
Is that correct?

hamishmack:
Yes that looks like the problem.  These constraints [here](https://github.com/cardano-foundation/cardano-wallet-agda/blob/912820dc8aec76b590ea53db0c74ae8121c44818/cabal.project#L26-L37):

```
constraints:
    base >= 4.18.2.0 && < 5

  , bech32 >= 1.1.7

  -- Cardano Node dependencies:
  , io-classes >= 1.4
  , io-classes -asserts

  , ouroboros-consensus-cardano ^>= 0.18
  , ouroboros-network ^>= 0.16.1
  , ouroboros-network-protocols ^>= 0.9
```

Are not visible to CHaP and it is testing the packages without them.  You can see the dependencies picked on [line 293 of the log](https://github.com/IntersectMBO/cardano-haskell-packages/actions/runs/13043668079/job/36390587302?pr=996#step:7:293).  They include things like `--dependency=ouroboros-consensus-cardano=ouroboros-consensus-cardano-0.21.0.0-2dmx1khfGo51sXXs7f9J0w` (instead of 0.18).

The best solution is probably to move those constraints from the `cabal.project` into the `.cabal` files and make a new PR.

I suspect you might still need to use the trick in this PR to disable testing with GHC 8.10.

abailly:
That makes sense. I completely forgot about those bounds, and the fact they were set at `cabal.project` level and not individual `.cabal` files which obviously is problematic when one wants to publish individual libraries 😅 
As for the 8.10.7 test execution issue, we could just bypass it the same way: Add a condition in the cabal file?

hamishmack:
> Add a condition in the cabal file?

Unfortunately I think CHaP does not know to look for `buildable` in the `.cabal` file and it might still try to build it (and fail at the planning stage).

PR #1039, opened 2025-03-26
cardano-ledger-binary/1.4.0.0: Bump upper bound on a dependency

erikd:
`quickcheck-instances`


erikd:
Bad idea. breaks build off a test lib in `ledger` due to a duplicate instance declaration (one in the `ledger` testlib and a new one in `quickcheck0instances-0.3.32`.

PR #1040, opened 2025-03-26
Disable macOS builds with GHC 8.10

hamishmack:
With recent versions of nixpkgs GHC 8.10 seems to be broken on macOS.

See https://gitlab.haskell.org/ghc/ghc/-/issues/25608


erikd (approved):
LGTM!

PR #1378, opened 2026-05-26
Revision ouroboros-consensus 3.0.1.0 for validation <1.2

jasagredo:

### DIFF
--- a/_sources/ouroboros-consensus/3.0.1.0/meta.toml
+++ b/_sources/ouroboros-consensus/3.0.1.0/meta.toml
@@ -1,2 +1,6 @@
 timestamp = 2026-04-14T21:25:56Z
 github = { repo = "IntersectMBO/ouroboros-consensus", rev = "c87aa760001e60f0f0d3353f793eb089adb917e7" }
+
+[[revisions]]
+number = 1
+timestamp = 2026-05-28T08:12:31Z
--- a/_sources/ouroboros-consensus/3.0.1.0/revisions/1.cabal
+++ b/_sources/ouroboros-consensus/3.0.1.0/revisions/1.cabal
@@ -1,0 +1,1997 @@
+cabal-version: 3.0
+name: ouroboros-consensus
+version: 3.0.1.0
+synopsis: Consensus layer for the Ouroboros blockchain protocol
+description: Consensus layer for the Ouroboros blockchain protocol.
+license: Apache-2.0
+license-files:
+  LICENSE
+  NOTICE
+
+copyright:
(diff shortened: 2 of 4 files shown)


neilmayhew:
@jasagredo I see you pushed a commit that changes the list of smoke test packages to include `ouroboros-consensus` instead of `ouroboros-consensus-cardano` but this hasn't solved the CI failures.

Based on the experiments I've been doing, I think what's needed is a revision to `ouroboros-consensus-cardano` that adds a constraint on `validation`. I've been able to nix-build `allSmokeTestPackages` successfully with that. However, I haven't tried it in CI. Would it be OK if I push a commit containing that to your branch?

neilmayhew:
Actually, I see that it's only Hydra that's failing, and I haven't been trying the same build that it's doing. That's probably failing for a different reason. It used to be that Hydra builds of CHaP always failed so we ignored them, but I tried hard to fix that. I'll start investigating that angle instead. My previous change may not be relevant after all.

neilmayhew:
The Hydra jobs fail because they're using the state of CHaP as it was *before* the change, so don't have the revised version of `ouroboros-consensus` with the constraint. This is because CHaP is a flake input with the url `github:input-output-hk/cardano-haskell-packages?ref=repo`.

The GitHub CI builds essentially the same things, also using nix, but uses `--override-input CHaP path:_repo` with a freshly-built repo in `_repo`. Hydra doesn't do this, and probably can't do it without a major rework of all the nix code, so Hydra builds will fail until the `repo` branch in the git repo is updated by CI after this PR is merged. However, if I build the same flake packages as Hydra, locally, but with the override, they succeed, so I'm confident that Hydra builds will start working again after the merge.

I think you should therefore ignore the Hydra failures and go ahead with merging.

However, I'm concerned that `ouroboros-consensus-cardano` isn't getting a revision, since some other CHaP packages may still have it as a dependency. If you think it's needed, and want to cherry-pick my commit, it's in fb1416f4666d5b6f754528adda915f653f7276d3.

jasagredo:
I have removed `ouroboros-consensus-Cardano` from the smoke test packages because that package is gone since `ouroboros-consensus 1.0.0.0`. We unified all packages n in to one, so the 0.26.0.3 it tried to build is ancient

neilmayhew (approved):

jasagredo:
Ah I rebased the branch but now maybe I need to also put a later timestamp. Let's see what CI says

jasagredo:
Aha indeed. Will fix

PR #1395, opened 2026-06-29
Added cardano-crypto-leios-0.1.0.0 and cardano-base-0.1.6.0.0

ch1bo:
Release of new package `cardano-crypto-leios`, which also needs `cardano-base >=0.1.6`


lehins (changes requested):
There is another PR that needs to be merged before we can release it: https://github.com/IntersectMBO/cardano-base/pull/675

lehins (approved):
👍

PR #1401, opened 2026-07-03
Add a script and CI check that new packages already exist on Hackage

neilmayhew:
If a package with the same name as one on CHaP is later added to Hackage, it prevents us from eventually migrating the CHaP package to Hackage. Our policy is to create an unpublished package candidate on Hackage for every package on CHaP that isn't already published to Hackage. This check ensures the invariant is maintained.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -184,6 +184,37 @@ jobs:
           --override-input CHaP path:_repo
           --show-trace
 
+  check-existence:
+    runs-on: ubuntu-latest
+    needs:
+      - build-repo
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Download package metadata
+        uses: actions/download-artifact@v8
+        with:
+          name: package-metadata
+          path: .
+
+      - name: Check that new package names also exist on Hackage
+        run: |
+          pkg-names() { jq -r 'map(."pkg-name") | unique | .[]' "$@"; }
+
+          MISSING=$(mktemp); trap 'rm -f "$MISSING"' EXIT
+          LANG=C join -v2 <(pkg-names packages-old.json) <(pkg-names packages-new.json) |
+            xargs -r scripts/not-on-hackage.sh >"$MISSING"
+          [ -s "$MISSING" ] || exit 0
+
+          # Output a multi-line annotation
+          sed '1s/^/::error::/;$!s/$/%0A/' <<EOF | tr -d '\n'; echo
+          The following packages do not exist on Hackage:
+          $(sed 's/^/* /' "$MISSING")
+          Please ensure they are published or are candidates before adding them to CHaP
+          EOF
+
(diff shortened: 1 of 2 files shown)


neilmayhew:
Haha! This already found two existing packages that don't exist on Hackage:

* `cardano-crypto-leios`
* `hermod-recon-framework`

lehins (commented):
> This already found two existing packages that don't exist on Hackage:

This is great. However it doesn't look like there is a failure on CI because of this, i.e. this job succeeded?

Before we can enforce this check on CI we need to provide helpers for creating such candidates on Hackage.

Here is a repository that contains initial work, which needs to be transferred to IntersectMBO: https://github.com/lehins/cardano-hackage-candidates 
That repository contains cabal files for all existing candidates and the initial script that was used to publish all existing packages as candidates. We need to adjust the script and add instructions on how to publish candidates on Hackage. @neilmayhew, could you please take care of the script and readme in the above repo, I gave you permissions to it.

@Ranchhand87, could I please ask you to create https://github.com/IntersectMBO/cardano-hackage-candidates repository and give all the same permissions to users/groups as to CHaP to that repo. Once we have access we'll transfer the contents of the repo to IntersectMBO repo.

neilmayhew:
> This is great. However it doesn't look like there is a failure on CI because of this, i.e. this job succeeded?

It's possible to check all CHaP packages every time, which is how I found this. However, it involves spamming Hackage with API calls that we already know the answer to. It also means that any existing packages that don't have candidates will cause CI to fail for everyone until they're fixed. I therefore chose to check only the packages that are being added in the PR that this is running for, and it passes for this PR because the PR doesn't add any new packages. (I did test it with a temporary commit that added a new, fake package.)

If you think it's better to check all packages every time, that's an easy change. The spamming occurs only for packages that don't have any candidates, ie that are already on Hackage as published packages, which currently is only a handful.

neilmayhew:
> We need to adjust the script and add instructions on how to publish candidates on Hackage. @neilmayhew, could you please take care of the script and readme in the above repo

Will do

neilmayhew:
One other point to note is that I chose to make this a separate job rather than making it part of the `build-new-packages` job. I did this so that the results would be more obvious, and so that the job could run in parallel with building packages, and could run on GitHub runners.

Because of this, if we want it to be able to fail the CI, it needs to be added to the required checks for the repo. I'm not an admin, so I can't do this myself.

neilmayhew:
As discussed, the plan is now to add the content of `lehins/cardano-hackage-candidates` to CHaP, and add some extra tooling around it.

However, I'll make that separate PR for that. I think this PR can be merged as-is. The only change needed after the other functionality is added will be to update the error message to include instructions on how to fix the problem (using the new functionality).

The check in this PR will flag new packages being added that aren't on Hackage, but it won't start failing the CI until the job is added to the set of required checks.

PR #1413, opened 2026-07-28
cardano-ledger release for node 11.1

f-f:
* byron-spec-chain-1.0.1.2
* byron-spec-ledger-1.1.0.2
* cardano-ledger-allegra-1.10.0.0
* cardano-ledger-alonzo-1.16.0.0
* cardano-ledger-api-1.14.0.0
* cardano-ledger-babbage-1.14.0.0
* cardano-ledger-binary-1.9.0.0
* cardano-ledger-conway-1.23.0.0
* cardano-ledger-core-1.21.0.0
* cardano-ledger-dijkstra-0.3.0.0
* cardano-ledger-mary-1.11.0.0
* cardano-ledger-shelley-1.19.0.0
* cardano-ledger-shelley-test-1.9.0.0
* cardano-protocol-0.1.0.0
* cardano-protocol-tpraos-1.6.0.0
* small-steps-1.1.4.0
* vector-map-1.2.1.0

From https://github.com/IntersectMBO/cardano-ledger at f649f9751074d2ab3de033fc3912f29c9862c1f5

### DIFF
--- a/_sources/byron-spec-chain/1.0.1.2/meta.toml
+++ b/_sources/byron-spec-chain/1.0.1.2/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2026-07-29T17:00:01Z
+github = { repo = "IntersectMBO/cardano-ledger", rev = "f649f9751074d2ab3de033fc3912f29c9862c1f5" }
+subdir = 'eras/byron/chain/executable-spec'
--- a/_sources/byron-spec-ledger/1.1.0.2/meta.toml
+++ b/_sources/byron-spec-ledger/1.1.0.2/meta.toml
@@ -1,0 +1,3 @@
+timestamp = 2026-07-29T17:00:01Z
+github = { repo = "IntersectMBO/cardano-ledger", rev = "f649f9751074d2ab3de033fc3912f29c9862c1f5" }
+subdir = 'eras/byron/ledger/executable-spec'
--- a/_sources/cardano-crypto-wrapper/1.3.0/meta.toml
+++ b/_sources/cardano-crypto-wrapper/1.3.0/meta.toml
@@ -5,3 +5,7 @@ subdir = 'eras/byron/crypto'
 [[revisions]]
 number = 1
 timestamp = 2023-03-21T14:44:35Z
+
+[[revisions]]
+number = 2
+timestamp = 2026-07-29T17:12:40Z
--- a/_sources/cardano-crypto-wrapper/1.3.0/revisions/2.cabal
+++ b/_sources/cardano-crypto-wrapper/1.3.0/revisions/2.cabal
@@ -1,0 +1,120 @@
+cabal-version: 2.2
+
+name:                cardano-crypto-wrapper
+version:             1.3.0
+synopsis:            Cryptographic primitives used in the Cardano project
+description:         Cryptographic primitives used in the Cardano project
+license:             Apache-2.0
+author:              IOHK
+maintainer:          operations@iohk.io
+copyright:           2019 IOHK
+category:            Currency
+build-type:          Simple
+extra-source-files:  README.md
+
+common base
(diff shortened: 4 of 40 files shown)


lehins:
@f-f Looks like we need to create a revision for every previously released versions of `cardano-crypto-wrapper` with a bound added: `cardano-crypto < 1.4`, due to this build failure:
```
       Last 25 log lines:
       >       In an equation for ‘prv’: prv = CC.generate seed pp
       >       In the expression:
       >         let prv = CC.generate seed pp in (CC.toXPub prv, prv)
       >    |
       > 16 |   let prv = CC.generate seed pp in (CC.toXPub prv, prv)
       >    |             ^^^^^^^^^^^
       > [20 of 22] Compiling Cardano.Crypto.Signing.KeyGen ( src/Cardano/Crypto/Signing/KeyGen.hs, 
```
which is caused by this release: https://github.com/IntersectMBO/cardano-haskell-packages/pull/1412

and this PR: https://github.com/IntersectMBO/cardano-crypto/pull/106/changes

f-f:
@lehins thanks for the context - I cut the revisions, but it looks like the issue is deeper than that: https://github.com/IntersectMBO/cardano-ledger/pull/5874 bumped `cardano-crypto-class` to allow `2.6`, which has `crypton ^>= 1.1`, which means we can't pick `cardano-crypto@1.3.0`, because it binds `crypton >= 0.32 && < 1.1`.
`cardano-crypto@1.2` is then picked, and it breaks since it has no upper bound on `crypton` and the new API is just different.

I suppose we can keep the train moving and get a new `crypto-wrapper` + fix the compat with the new `ram` package downstream? (i.e. https://github.com/IntersectMBO/cardano-ledger/pull/5956 + plutus compat + whatever else that breaks downstream)

lehins:
> https://github.com/IntersectMBO/cardano-ledger/pull/5874 bumped cardano-crypto-class to allow 2.6

:sweat_smile: Why did you update the bound to include 2.6? That version of cardano-crypto-class was the whole reason why you had to close this PR: https://github.com/IntersectMBO/cardano-haskell-packages/pull/1411

This is what I was talking about in slack:

> Important piece of info for Ledger+ram compatibility. SRP for CLS needs an update to include this https://github.com/tweag/cardano-cls/pull/310 (edited) 
> You know, I take it back. I suspect it will take a while to upgrade to ram, since other repos also depend on memory.


So, there are two paths forward. Feel free to pick the one that you like best:
* Make all packages compatible with `ram`. Here is a WIP from Erik for Ledger: https://github.com/IntersectMBO/cardano-ledger/pull/5956 
* Add back an upper bound on <2.7 in ledger packages and deal with `ram`+`crypton-1.1` upgrade with a later release

Naturally, former one will require more work than the latter

f-f:
> Why did you update the bound to include 2.6?

It was in the changeset from Joosep and I didn't realise it was there 🙂
I will try to lower the bound and see if everything is still happy.

lehins:
> It was in the changeset from Joosep and I didn't realise it was there 🙂

I see. Makes sense. I just looked up a commit that was made by you and assumed that you introduced this upper bound: https://github.com/IntersectMBO/cardano-ledger/pull/5874/changes/70aa865717157bc8fe4a58646161f64ceb311524

It took me a moment, but I remember now. That was definitely a mistake in the above PR into Ledger. That upper bound should not have been relaxed.

What confused me was that `cardano-crypto-class-2.6.0.0` was barely prepared and released a few days ago, which is why I didn't even think that it was Joosep who relaxed this upper bound, since he was on PTO during all that time :smile: 
But then I remembered that he tried to introduce a breaking change and temporarily there was a `cardano-crypto-class-2.6.0.0` in a his PR into cardano-base: https://github.com/IntersectMBO/cardano-base/pull/665#discussion_r3543894889 :sweat_smile:

lehins (approved):

PR #1414, opened 2026-07-29
Fixed add-from-github.sh script

coot:
This PR fixes two problems:
* the script failed with exit code `-1`, when there is no hackage candidate.
* `NEW_CANDIDATES` is empty, if the script `./hackage-candidates/create-placeholders.sh` is run without `--no-commit` flag, which results in a missing warning message.

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -184,7 +184,7 @@ rm -rf "$WORKDIR"
 
 git log -1 --name-status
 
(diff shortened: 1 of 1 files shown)


neilmayhew:
Thanks for catching and fixing this. Sorry for the trouble caused by my hasty changes made just before going on vacation!

Some thoughts:

* It's better to emit the warning in the `create-placeholders` script, at the time the candidate is created, rather than later when the rest of the package is added.

* The problem with grep can be handled more elegantly by not using grep at all.

I've created a PR (#1417) and added you as a reviewer.

PR #1417, opened 2026-07-29
Improve handling of new Hackage candidates in scripts

neilmayhew:
This is a follow-up to #1414

### DIFF
--- a/scripts/add-from-github.sh
+++ b/scripts/add-from-github.sh
@@ -182,9 +182,9 @@ fi
 log "Removing work directory $WORKDIR"
 rm -rf "$WORKDIR"
 
-git log -1 --name-status
+git log --name-status @^..
 
-NEW_CANDIDATES=$(git log --name-only --format= main.. | { grep ^hackage-candidates/ || true; } | xargs sed -n '/^name: */Is///p')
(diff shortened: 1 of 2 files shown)


scripts/add-from-github.sh
```diff
@@ -182,9 +182,9 @@ fi
 log "Removing work directory $WORKDIR"
 rm -rf "$WORKDIR"
 
-git log -1 --name-status
+git log --name-status @^..
 
-NEW_CANDIDATES=$(git log --name-only --format= main.. | { grep ^hackage-candidates/ || true; } | xargs sed -n '/^name: */Is///p')
+NEW_CANDIDATES=$(git log --name-only --format= @^.. hackage-candidates/ | xargs -r -d'\n' sed -n '/^name: */Is///p')
```
coot, line 187: I prefer `main..` over `@^..`

neilmayhew: `main..` will warn about any commits on this branch, even if they weren't made by this run of the script. For example, if `add-from-github.sh` was run previously to add a new package, and then run again to add a new version of an existing package, a warning will be generated by both runs of the script. An extra warning isn't a bad thing in itself, but it could be confusing if the user wasn't expecting to add a new package with this run of the script.

coot (approved):

PR #1431, opened 2026-08-11
Added cardano-cli-11.2.0.0

Jimbo4350:
From https://github.com/IntersectMBO/cardano-cli at 4aa8f5b33ff4a7b54025680b12df59a838b8dcf5 (tag `cardano-cli-11.2.0.0`)

Release PR: https://github.com/IntersectMBO/cardano-cli/pull/1416


Jimbo4350:
Updated to the re-cut `cardano-cli-11.2.0.0` tag (`be6bd1c06`, from IntersectMBO/cardano-cli#1416). The original sdist was unbuildable with default cabal flags because `cardano-cli` used `cardano-diffusion:ping`'s command line parser, which requires cardano-diffusion's manual `optparse-applicative-fork` flag (set via `cabal.project`, which doesn't ship with the sdist). That's fixed by IntersectMBO/cardano-cli#1413 — the parser is now local to cardano-cli and behaviourally identical, so this package builds with default flags. The old tag was deleted and re-pointed before anything was published.

PR #1452, opened 2026-09-04
Add cardano-api-11.7.0.0

carbolymer:
From https://github.com/IntersectMBO/cardano-api tag cardano-api-11.7.0.0


palas (changes requested):
It looks like the commit hash is wrong. Maybe it changed with the signing

palas (approved):
Clarified, it is a hash to a tag, I didn't know those existed

PR #1456, opened 2026-09-08
Added cardano-slotting-0.2.2.0

gdeest:
From https://github.com/IntersectMBO/cardano-base at b6559d4a1eb6aca039612a5b3d908590727d31ff


lehins (approved):
Needs an update to timestamp, since another PR was merged, but otherwise it is good to go.

